PR #59, opened 2020-08-28
use shared client for throttling per host

### LINKED ISSUE #45: Throttle # of connections to api
When Ajour performs a refresh or update all operation, a large number of simultaneous asynchronous calls are being made to the API (curse, tukui, wowinterface). I suggest a setting that caps/throttles the number of connections.

I do not think this value needs to be exposed in the config file / settings view as most users will probably see it as a way to "optimize" their performance and not consider the API. But if we do expose it, it should only allow a range of sensible values.

For reference, a typical value used by browsers is 6 per server.

See Firefox about.config setting: network.http.max-persistent-connections-per-server

tarkah:
resolves #45 

Haven't tested this as I don't have any WoW installed / addons on my machine. It should limit 6 connections per API host by using a shared client across all requests and specifying `max_connections_per_host(6)`

### DIFF
--- a/src/gui/mod.rs
+++ b/src/gui/mod.rs
@@ -9,9 +9,14 @@ use crate::{
     error::ClientError,
     tukui_api, wowinterface_api, Result,
 };
+use async_std::sync::Arc;
 use iced::{
     button, pick_list, scrollable, Application, Column, Command, Container, Element, Length,
     Settings, Space,
+};
+use isahc::{
+    config::{Configurable, RedirectPolicy},
+    HttpClient,
 };
 use std::path::PathBuf;
 
@@ -64,6 +69,7 @@ pub struct Ajour {
     config: Config,
     expanded_addon: Option<Addon>,
     is_showing_settings: bool,
+    shared_client: Arc<HttpClient>,
 }
 
 impl Default for Ajour {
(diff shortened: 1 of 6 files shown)


src/gui/mod.rs
```diff
@@ -80,6 +86,13 @@ impl Default for Ajour {
             config: Config::default(),
             expanded_addon: None,
             is_showing_settings: false,
+            shared_client: Arc::new(
+                HttpClient::builder()
+                    .redirect_policy(RedirectPolicy::Follow)
+                    .max_connections_per_host(6)
+                    .build()
+                    .unwrap(),
```
tarkah, line 94: Program can panic here, not sure how you want to handle this inside `<Ajour as Application>::new`

casperstorm:
Hi @tarkah.
I've tested this and it works great. This is a clever way of solving this, and is much nicer against the different API's.
Thank you so much for contributing.

src/gui/mod.rs
casperstorm: Good question. For now I think its okay to continue and tackle this task while doing improvements later.

tarkah:
Great, glad it worked out! I stumbled across this project today and it's awesome. Wish this existed back when I was still playing classic.

casperstorm:
> Great, glad it worked out! I stumbled across this project today and it's awesome. Wish this existed back when I was still playing classic.

I am so glad you like it, and even more glad you are helping contribute!

PR #60, opened 2020-08-29
make columns sortable

### LINKED ISSUE #37: Group addons with updates at top of list
It's not ideal having to scroll through a long list to look for updates. The UX could benefit from grouping addons that have updates at the top of the list when a refresh finishes. Preferably before the user is able to interact with the list.

tarkah:
resolves #37 

Fix `Ord` and `PartialOrd` to sort properly. Updatable addons sort to top, then sort by title ascending. If title's match, then sort by remote version descending.

Sort addons after folder is initially parsed, each addon metadata is downloaded (to determine if updatabable or not) and after each package is updated (to resort by name).

### DIFF
--- a/src/gui/update.rs
+++ b/src/gui/update.rs
@@ -610,30 +610,53 @@ async fn perform_unpack_addon(
 }
 
 fn sort_addons(addons: &mut [Addon], sort_direction: SortDirection, sort_key: SortKey) {
-    match sort_key {
-        SortKey::Title => {
+    match (sort_key, sort_direction) {
+        (SortKey::Title, SortDirection::Asc) => {
             addons.sort();
         }
-        SortKey::LocalVersion => {
+        (SortKey::Title, SortDirection::Desc) => {
+            addons.sort_by(|a, b| {
+                a.title
+                    .cmp(&b.title)
+                    .reverse()
+                    .then_with(|| a.remote_version.cmp(&b.remote_version))
+            });
+        }
+        (SortKey::LocalVersion, SortDirection::Asc) => {
             addons.sort_by(|a, b| {
                 a.version
                     .cmp(&b.version)
                     .then_with(|| a.title.cmp(&b.title))
             });
         }
-        SortKey::RemoteVersion => {
+        (SortKey::LocalVersion, SortDirection::Desc) => {
+            addons.sort_by(|a, b| {
+                a.version
+                    .cmp(&b.version)
+                    .reverse()
+                    .then_with(|| a.title.cmp(&b.title))
+            });
+        }
+        (SortKey::RemoteVersion, SortDirection::Asc) => {
             addons.sort_by(|a, b| {
                 a.remote_version
                     .cmp(&b.remote_version)
                     .then_with(|| a.cmp(&b))
             });
         }
-        SortKey::Status => {
+        (SortKey::RemoteVersion, SortDirection::Desc) => {
+            addons.sort_by(|a, b| {
+                a.remote_version
+                    .cmp(&b.remote_version)
+                    .reverse()
+                    .then_with(|| a.cmp(&b))
+            });
+        }
+        (SortKey::Status, SortDirection::Asc) => {
             addons.sort_by(|a, b| a.state.cmp(&b.state).then_with(|| a.cmp(&b)));
         }
-    }
-
-    if sort_direction == SortDirection::Desc {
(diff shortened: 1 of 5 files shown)


src/gui/update.rs
```diff
@@ -145,6 +145,8 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
                     // Update the addon with the newly parsed information.
                     addon.update_addon(updated_addon);
                 }
+
+                ajour.addons.sort();
```
tarkah, line 149: Remove this if you want to keep addons that were just updated at the top of the list until 'Refresh' is explicitly clicked.

tarkah:
Resolves #37

casperstorm:
Hi @tarkah, thanks for this contribution!
This is tested, and it works. However the problem is that calling `ajour.addons.sort();` many times causes the application to become extremely laggy. I don't know if it's `iced` or what, but there is a performance hit for sure.

I would love to tackle this performance hit before merging so that it is running smooth while sorting. Are you able to see the performance hit as well? Currently I am running my tests on 50-60 addons.

tarkah:
I didn't notice a performance hit, but I only tested with ~10 add-ons. Yeah, it's sorting the entire vec after every add-on is updated against, so very inefficient. We need a way to trigger the sort only after the last add-on is refreshed.

casperstorm:
> I didn't notice a performance hit, but I only tested with ~10 add-ons. Yeah, it's sorting the entire vec after every add-on is updated against, so very inefficient. We need a way to trigger the sort only after the last add-on is refreshed.

Yeah. As I see it the right way, for the user, is to sort every time an addon is updated, like you did - because then they see it immediately. But then we have the performance issue.
If we sort after all addons are done refreshing we might end up in a situation where the user thinks everything is good, and THEN it sorts. Causing sort to happen after a delay, which is not that UX friendly. It's hard to get right I think 🤔

mlablah:
Just my two cents on this matter.

When the list is being refreshed there is a status message informing the user that something is being worked on. "Fetching data for X addons" where the number X gradually decreases to 0. So the user is aware that the app is working on refreshing the list. IMO this primes the user to expect a late reordering and doing so would make the list less "jumpy".

On the other hand, when an update all is initiated, there isn't a status message signalling when the entire operation is done. Assuming addons with updates have already been sorted to the top, the user is able to tell from their status when the operation is done. In this scenario, reordering after every update or only reordering after the last update seems equally valid. There is another option though, and that is never reordering when doing an update (!). This way the user can see the new status of the recently updated addons the same place in the UI as before clicking update all / update. 

Ideas for new enhancement issues:
1) Status message when updating addons. Something like: "Updating all addons. X addons remaining."
2) Allow user to resort list by clicking the Addon column title. Less work than a full refresh that also has to parse folders and retrieve metadata again.

casperstorm:
How do you feel about this one @tarkah?
I can't really come up with any great solution right now.

tarkah:
I think the easy option here is just rely on sort-able columns and leave it up to the user. If they sort on status, we will have `Updatable` order first. And only allow sorting on headers once initial load, "Refresh" or "Update All" operations have fully completed. That way the user understands exactly what is happening and sorting only happens once when the header is clicked, so performance won't be impacted.

casperstorm:
Think that sounds reasonable. You are very welcome to give it a shot. Let me know if I can help with anything.

tarkah:
@casperstorm Ok, check out the latest commit. I've made the columns sortable (changed the headers to buttons that trigger a sort to occur).

It's worth noting Title and Local version sorts are persistent across clicking "Refresh" since those attributes live on the file system. Any sort on Remote version and Status will be lost when clicking "Refresh" since both those attributes are lost once the file system is parsed and the sorting occurs before all the async operations against the API. In these instances, it'll just resort based on Title.

Please confirm the styling is ok. I was going to make the entire column width click able, but with the left spacer in place, it didn't look right.

casperstorm:
@tarkah - Impressive 👏🏻 I really like this!
Coding is clean as well, nothing to put my finger on.

I think we can tweak the headers a bit, but that is something we can adjust further down the road – I don't want to hold this PR back just because of pixel pushing, to be honest.

This is a **GREAT** addition!

tarkah:
Awesome! Thanks for letting me tackle these, it's been fun :)

PR #69, opened 2020-09-02
fix: line height

### LINKED ISSUE #68: Rendering artifacts in Local version text
I'm getting rendering artifacts in the Local version text. 

image

casperstorm:
This should fix #68.
Do you mind testing this @tarkah?

### DIFF
--- a/src/gui/element.rs
+++ b/src/gui/element.rs
@@ -167,7 +167,7 @@ pub fn settings_container<'a>(
 }
 
 pub fn addon_data_cell(addon: &'_ mut Addon, is_addon_expanded: bool) -> Container<'_, Message> {
-    let default_height = Length::Units(28);
+    let default_height = Length::Units(25);
 
     // Check if current addon is expanded.
     let version = addon.version.clone().unwrap_or_else(|| String::from("-"));


tarkah:
Yup, that fixes it! Though, it'd be nice to make the rows bigger if needed and not run into this issue. I am not familiar with the standards of the version string, why do some of mine have text on a new line after the first line? Maybe we can only preserve the first line of the installed version string?

casperstorm:
Yeah, its a nice fix. Let me explain my thoughts here. I can't control string truncation in `iced` as far as i know – so by keeping the row height equal or less than one line height will make sure we dont render some of the next line.

I've set the local version to a fixed width, so all cells have the same width:
https://github.com/casperstorm/ajour/blob/e0ab1466abab2a3f445957f265b11bddb55ed76c/src/gui/element.rs#L203
Removing this will cause the UI to be a bit cluttered because all local version containers will have different widths:

Screenshot 2020-09-02 at 18 36 05

I am open for suggestions here if we can do something smart.

tarkah:
I was thinking more in terms of bigger in height, not width. It would suck not to be able to make the rows taller in the future, just because the version string is multi-line and the second line starts showing.

How about before we save the version string to the `Addon`, we do something like `version.lines().next().unwrap_or("-")`. That way we guarantee it's only one line of text and rendering it inside a cell of any height will be safe

casperstorm:
Heres the UI without the spacing – then we could have the version containers grow as much as they would. But again i think i like the "spaced" UI more, or is that just me?
Screenshot 2020-09-02 at 18 44 48

casperstorm:
> I was thinking more in terms of bigger in height, not width. It would suck not to be able to make the rows taller in the future, just because the version string is multi-line and the second line starts showing.
> 
> How about before we save the version string to the `Addon`, we do something like `version.lines().next().unwrap_or("-")`. That way we guarantee it's only one line of text and rendering it inside a cell of any height will be safe

Ah, good idea – I'll look into that.

tarkah:
@casperstorm I just tried this and it doesn't work, since `\n` only counts as a single character, not the entire width of the line when you put it into the iced layout. We should split on lines (`\n`) and keep only the first line.

casperstorm:
@tarkah – could you take a look at this solution?
Idea is to have (for now) `readable_local_version` and `readable_remote_version` which is truncated to our desire.
I save it on `Addon` to avoid overhead on every view call.

tarkah:
@casperstorm I just tried this and it doesn't work, since `\n` only counts as a single character, not the entire width of the line when you put it into the iced layout. We should split on lines (`\n`) and keep only the first line.

image

casperstorm:
> @casperstorm I just tried this and it doesn't work, since `\n` only counts as a single character, not the entire width of the line when you put it into the iced layout. We should split on lines (`\n`) and keep only the first line.
> 

@tarkah – I just pushed an addition where I split on `\n` and then truncates it as well, just in case. Let me know if it works for you. I can't get it to break on macOS at least 🙂

tarkah:
image

Hmmm, still no luck. I guess there is no newline in there, it must be that its getting text wrapped. Is there a way in `iced` to disable text wrapping?

casperstorm:
> 
> 
> Hmmm, still no luck. I guess there is no newline in there, it must be that its getting text wrapped. Is there a way in `iced` to disable text wrapping?

What! Can you share the .toc file for one of the addons which fails? I've tried with a very very long line and it works on both my Mac and PC so I'll debug it a little tonight.

And no there's not. Talked to the owner yesterday and there's really no good way of solving this 😞

casperstorm:
@tarkah I am not happy with this solution.
First of all, it makes me nervous it's working for me, but not you. Second, this is a thing the GUI framework should do for us. Either capping a label to n amount of lines, or respect width.

If one of us comes up with a better solution than what we have now (capping cell height) then let's look at this issue again. For now i'll close this. I have pushed #72 which fixes it.

tarkah:
Yeah, absolutely agreed. As I commented on the issue, we're out of luck until the `iced` gives us a way to disable text wrapping. Your other PR is more than adequate for now. If we really need taller rows in the future, we can probably get away with using additional padding to keep the 2nd line from showing up.

PR #70, opened 2020-09-02
add theme feature

### LINKED ISSUE #65: Support light theme
This is a low priority task, but it would be easy to extend the curent `ColorPalette ` to support a light theme:

https://github.com/casperstorm/ajour/blob/d765c4675797c09bbac5af7e70c51b78f7edd991/src/gui/style.rs#L12-L23

An example of light theme colors could be: https://coolors.co/6100ee-00dac6-000000-f5f5f5-af0120-ffffff:

```rust
ColorPalette::Primary => Color::from_rgb(0.39, 0.0, 0.93),
ColorPalette::Secondary => Color::from_rgb(0.0, 0.85, 0.77),
ColorPalette::Surface => Color::from_rgb(0.96, 0.96, 0.96),
ColorPalette::OnSurface => Color::from_rgb(0.0, 0.0, 0.0),
ColorPalette::Background => Color::from_rgb(1.0, 1.0, 1.0),
ColorPalette::Error => Color::from_rgb(0.68, 0.0, 0.12),
```
Essentially the value should be saved to the `Config` and changeable though Settings. 

Screenshot 2020-08-31 at 21 50 33

tarkah:
will resolve #65 

Implemented so far:

- New `Theme` struct which will hold Colors
- Store theme in app and pass to styling to use colors of theme
- Custom deserializer for `iced::Color` to deser from hex string
  such as '#09ACDF'
- Only use windows subsystem on --release so we can run tests in debug
  mode

### DIFF
--- a/src/theme/mod.rs
+++ b/src/theme/mod.rs
@@ -1,6 +1,11 @@
+use crate::fs;
 use de::deserialize_color_hex_string;
 use serde::Deserialize;
 use std::cmp::Ordering;
+
+pub async fn load_user_themes() -> Vec<Theme> {
+    fs::load_user_themes().await
+}
 
 #[derive(Debug, Clone, Deserialize)]
 pub struct Theme {
@@ -63,6 +68,14 @@ impl PartialEq for Theme {
 impl PartialOrd for Theme {
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
         Some(self.name.cmp(&other.name))
+    }
+}
+
+impl Eq for Theme {}
+
+impl Ord for Theme {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.name.cmp(&other.name)
     }
 }
 
--- a/src/theme/mod.rs
+++ b/src/theme/mod.rs
@@ -1,0 +1,107 @@
+use de::deserialize_color_hex_string;
+use serde::Deserialize;
+
+#[derive(Clone, Copy, Deserialize)]
+pub struct Theme {
+    #[serde(deserialize_with = "deserialize_color_hex_string")]
+    pub primary: iced::Color,
+    #[serde(deserialize_with = "deserialize_color_hex_string")]
+    pub secondary: iced::Color,
+    #[serde(deserialize_with = "deserialize_color_hex_string")]
+    pub surface: iced::Color,
+    #[serde(deserialize_with = "deserialize_color_hex_string")]
+    pub on_surface: iced::Color,
+    #[serde(deserialize_with = "deserialize_color_hex_string")]
+    pub background: iced::Color,
+    #[serde(deserialize_with = "deserialize_color_hex_string")]
+    pub error: iced::Color,
+}
+
+impl Default for Theme {
+    fn default() -> Self {
+        Theme {
+            primary: iced::Color::from_rgb(0.73, 0.52, 0.99),
+            secondary: iced::Color::from_rgb(0.88, 0.74, 0.28),
+            surface: iced::Color::from_rgb(0.12, 0.12, 0.12),
+            on_surface: iced::Color::from_rgb(0.88, 0.88, 0.88),
+            background: iced::Color::from_rgb(0.07, 0.07, 0.07),
+            error: iced::Color::from_rgb(0.76, 0.19, 0.28),
+        }
+    }
+}
+
+// Newtype on iced::Color so we can impl Deserialzer for it
+struct Color(iced::Color);
+
+mod de {
+    use super::Color;
+    use serde::de::{self, Error, Unexpected, Visitor};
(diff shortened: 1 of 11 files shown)


tarkah:
I've added a theme picker to the application under settings, fairly identical to the flavor picker.

Note I chose BTreeMap to store name of Theme as key and Theme as value. I chose this to preserve the order in which we insert to this, since I will always insert the static themes Dark and Light before any user themes. I plan on inserting user themes in alphabetical order after the Dark / Light themes.

src/theme/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+
+// Newtype on iced::Color so we can impl Deserialzer for it
+struct Color(iced::Color);
+
+mod de {
+    use super::Color;
+    use serde::de::{self, Error, Unexpected, Visitor};
+    use std::fmt;
+
+    pub fn deserialize_color_hex_string<'de, D>(deserializer: D) -> Result<iced::Color, D::Error>
```
casperstorm, line 48: Really awesome job on this deserializer, @tarkah.

src/theme/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+                Err(de::Error::invalid_value(Unexpected::Str(s), &self))
+            }
+        }
+
+        deserializer.deserialize_any(ColorVisitor).map(|c| c.0)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
casperstorm, line 91: Yay 🎉
You've implemented our first tests.

casperstorm (approved):

casperstorm:
This is is already a great addition 👏🏻
I've reviewed this part now and it looks good. You can continue to push commits to this if it make sense, then i'll just review as we go.

tarkah:
Awesome! Yup, I'll just push out pieces as I complete them. I mainly need to add the theme folder support, which has a lot of redundancy with identifying the primary and secondary location as the config file, so I may refactor that slightly so I'm not having to duplicate any code there.

casperstorm:
> Awesome! Yup, I'll just push out pieces as I complete them. I mainly need to add the theme folder support, which has a lot of redundancy with identifying the primary and secondary location as the config file, so I may refactor that slightly so I'm not having to duplicate any code there.

Speaking of multiple config file location. In the beginning the config file was not created by the application, which is why we allowed to store the config file multiple places, depending on what the user liked. This, however, is strictly speaking not necessary anymore because we have the settings view and really don't need a user to edit the config. So, if you are refactoring and it can help out I think we can commit to only allow the config (and themes) to be located at a specific location:

```rust
// macOS/Linux
$HOME/.config/ajour/ajour.yml
// Windows
%APPDATA%\ajour\ajour.yml
```

casperstorm:
Also, a suggestion is to convert this PR to a draft until it is ready to be merged. Then we can do review and talk in here without confusing the other PR's which is waiting.

tarkah:
I've updated `Theme` to include a name, since I don't want to rely on the file name to determine what shows up in the pick-list for the user. It should be more human readable than that. Because of this, `Theme` no longer implements `Copy` and so I moved the color definitions to a `ColorPalette` struct inside `Theme` that does implement `Copy`. Since we need to store the `ColorPalette` inside each styled element, it's very important it implements `Copy`.

I've also rebased.

casperstorm:
> I've updated `Theme` to include a name

Sounds good, this way we can also easier distinguish themes if you have a lot. Like Gruvbox, Nord, Solarized etc.

tarkah:
Alright, I think this is good to go! I added a function to asynchronously load all user configs from the filesystem when the app is loaded. These will show up in the picklist. If a user chooses something from the picklist, it will save that theme name to the config file and that will be used whenever the config is loaded into the application.

As a safety precaution, if the saved theme name in the config doesn't exist in the user configs returned, it'll default to using the Dark theme.

tarkah:
Are you happy with the Light theme? I never messed with it from what you originally defined.

tarkah:
Also, I'm using `async_std::fs` here, we should probably convert all the other filesystem operations over in a separate PR so they can't block the Iced loop for best user experience.

casperstorm:
> Alright, I think this is good to go!
Wooh! 🎉
I'll test right away.

> Are you happy with the Light theme? I never messed with it from what you originally defined.

I'll look into improving the light theme a little.

casperstorm (approved):
Tested. This is FANTASTIC.
Note: one thing we should do is update README and CHANGELOG with this change as well @tarkah.
Then later we will make a Wiki section with some example of different themes.

tarkah:
Will do! Can you add a theme page to the Wiki so I can at least reference it from the README? We can put Work in Progress for now until we fill it out.

PR #73, opened 2020-09-04
use PersistentData trait

tarkah:
I have no idea if there will ever be persistent data outside of the configuration file, but I've added a trait to handle loading / saving any serializable struct that needs to be persistent.

Basically, all you need to do now `impl PersistentData` onto your struct and define the `relative_path` you want it located at (relative to Ajour's config directory for each platform). It then exposes methods such as `load`, `load_or_default` and `save`.  `load_or_default` can be called if the struct is also `Default` and if the file doesn't already exist, it will save it and load it as `Default`.

This cleans up the `config` module so it's just defining the data structure, then the implementation detail of `PersistentData` in `fs` handles all the file system operations.

Can you test on Mac? I've confirmed it works on Windows.


casperstorm (approved):
This is really clean, @tarkah.
I **love** to go back and do these kind of optimizations because it means we have become wiser while at the same time improving the application for future us. If that make sense, haha.

As I also wrote earlier, I think it is fine to only have one path (to the config) for each platform. It was a relic of the past.

Let's continue to improve areas like this! GREAT job 👏🏻

PR #78, opened 2020-09-06
feat/fingerprint

### LINKED ISSUE #75: Fingerprint local addons to improve addon parsing
Currently our ability to match a local addon with an online repository is entirely dependent on the quality of the data in the local toc files. In many cases the `version` tag is missing from the toc file or it has an incorrect value.

To mitigate this, we can implement a fingerprint algorithm that will match the fingerprint values returned by the curse api's fingerprint endpoint.

This has implications for:
- [x] Addon parsing (need to calculate fingerprints)
- [x] GUI (show/hide addons based on additional criteria)
- [x] Updating (try fingerprint lookup before ever searching by title)

**Additional context**
Caching fingerprints per folder along with a timestamp, and only recalculate the fingerprint if the toc file is newer than the timestamp (or addon is updated via ajour) could speed up subsequent load times.

Ability to "fail safe" to our offline logic, in case the fingerprint algorithm is changed on us, and nothing matches or the api is offline.

Implementing this enhancement will most likely solve #67 and #74.

casperstorm:
Resolves #75.
This PR aims to refactor a large amount in the way we parse and match addons.


tarkah:
@casperstorm I've added the generic function for sending a POST request with Json data in the body. See my branch here, which currently diffs to the HEAD of this branch:

https://github.com/tarkah/ajour/tree/fingerprint-request

Since `Vec<u32>` is serializable, we can serialize it as a valid Json byte vector using `serde_json`. I've also added the headers and timeout options we have in our generic GET request function. As long as the data you pass this function is `Serialize`, it will work.

casperstorm:
Some things you can help me sort out, or answer @tarkah when you have time 🙂
1. I dont feel like this fits into this file, and I am a little unsure about naming as well. I am not sure where it should live though.
https://github.com/casperstorm/ajour/blob/8a9b618d80d022a0eac0d80e9635d7acab29683b/src/parse.rs#L22-L45

2. I need some regexes which is coming from the API right now: `fetch_game_info`. Problem is that I need them every time I do something with fingerprinting. We should store this somewere and make sure we only fetches it once per lifetime of the application. Maybe you can help refactor this somewhere clever? Or has an idea.
https://github.com/casperstorm/ajour/blob/8a9b618d80d022a0eac0d80e9635d7acab29683b/src/parse.rs#L51-L83

3. Clippy is complaining.

casperstorm:
@tarkah it is almost ready, the majority of the logic is in place. I need to add a loading screen/indicator but that is about it. Then I need to do a lot of testing (we should add tests...) but other than that you can review if you like!

tarkah:
@casperstorm I haven't had a chance to fully look through everything, but I took a stab at some of the items you brought up.

1. Currently all fingerprint related stuff is local to the `parse` module. I even removed the `pub` from those few things. So I think it's fine for it to live in there right now. 

2. Have you heard of `lazy_static`? I've used it here to have a globally mutable variable that we can store `GameInfo` into after the first API request. Therefore we can fetch it from memory to rebuild the Regex patterns instead of grabbing it from the API again. We could obviously store `GameData` within `Ajour` and pass it into the operation, but I was lazy and `lazy_static` let's you be exactly that :)

https://github.com/tarkah/ajour/blob/9831afad0e33db67a4f089369cd37c340f49c6d1/src/parse.rs#L23-L25

https://github.com/tarkah/ajour/blob/9831afad0e33db67a4f089369cd37c340f49c6d1/src/parse.rs#L53-L66

3. I've fixed the lints.

A couple other things to note:

- I changed `FingerprintCollection` into a type since it doesn't need to be anything more than a `Vec`. This will break your `fingerprints.yml` deserialization, so make sure to delete that file.

https://github.com/tarkah/ajour/blob/9831afad0e33db67a4f089369cd37c340f49c6d1/src/parse.rs#L33-L39

- I've made a new struct `ParsingPatterns` for the return type of `file_parsing_regex` since `(Regex, Regex, HashMap<String, (regex::Regex, Regex)>)` is a little rough on the eyes. You can de-structure this the same way as tuples, so it doesn't really change anything later on:

https://github.com/tarkah/ajour/blob/9831afad0e33db67a4f089369cd37c340f49c6d1/src/parse.rs#L123-L127

- We didn't need a new allocation here since we are just mutating the collection then re-saving it. `for_each` is great for this in lieu of a for loop.

```diff
@@ -309,22 +320,12 @@ pub async fn update_addon_fingerprint(addon: Addon) -> Result<()> {
         &extra_inclusion_regex,
         &file_parsing_regex,
     ) {
-        let new_collection: Vec<Fingerprint> = fingerprint_collection
-            .fingerprints
-            .clone()
-            .iter_mut()
-            .map(|fingerprint| {
-                if fingerprint.title == addon.id {
-                    fingerprint.hash = Some(hash);
-                }
-
-                fingerprint.clone()
-            })
-            .collect();
+        fingerprint_collection.iter_mut().for_each(|fingerprint| {
+            if fingerprint.title == addon.id {
+                fingerprint.hash = Some(hash);
+            }
+        });

-        let fingerprint_collection: FingerprintCollection = FingerprintCollection {
-            fingerprints: new_collection,
-        };
         let _ = fingerprint_collection.save();
```

casperstorm:
It feels like Christmas when I wake up, and see you have contributed, @tarkah 🎅🏻. Thank you for the additions.

> Currently all fingerprint related stuff is local to the parse module. I even removed the pub from those few things. So I think it's fine for it to live in there right now.

Great. I think so too, I just wanted a second opinion on this.

> Have you heard of lazy_static? [...] I understand if you don't want global variables, as it can get out of hand if abused, so we can always change this if needed.

I have heard about it, but then happily forgot about it again. Global variables **is** a heated topic to discuss, but as long as we use them correct, they do have their strengths so I like this solution.

> I changed FingerprintCollection into a type since it doesn't need to be anything more than a Vec. 

Writing the old `Struct` felt wrong, so thanks for changing this.

> I've made a new struct ParsingPatterns for the return type of file_parsing_regex since (Regex, Regex, HashMap<String, (regex::Regex, Regex)>) is a little rough on the eyes.

Haha, you are right about this one for sure 😄 

> We didn't need a new allocation here since we are just mutating the collection then re-saving it. for_each is great for this in lieu of a for loop.

Really good catch.

tarkah:
Glad the changes worked out! I like the addition of the loading indicator. Now you just need a super fancy animation to display for the indicator :P

PR #86, opened 2020-09-10
synchronize updates to FingerprintCollection

tarkah:
store FingerprintCollection in memory and only allow one operation on it at a time, so concurrent tasks can't overwrite changes to the collection. We still write changes to disk after each operation makes it's updates.


tarkah:
@casperstorm Test this out! I haven't tested yet, but I suspect it will work.

casperstorm:
This did the job, it seems!

tarkah:
Wooo! 🎉

PR #89, opened 2020-09-10
Optimizations

### LINKED ISSUE #87: Optimize parse module
@tarkah pointed out that the parse module is right now doing a lot of clones, and additional allocations.
We should eventually optimize this if possible.

### LINKED ISSUE #88: Addons not getting parsed after update through app
**Describe the bug**
I wanted to fully try out all the changes we implemented so I created a new addons folder with the top 16 WoW classic addons from curse, 13 of which I grabbed older versions to test the update functionality.

Launching the app, they all showed up very quickly! Woo, the app is fast and fetches the remote data quickly as well :) I then clicked Update All and the statuses all changed and updates took place very quickly, sweet!

However, once I clicked "Refresh" and things got re-parsed, instead of seeing 16 addons I now only see 13... I confirmed the addon folders contain the newly updated data (checked the TOC files).

I'll dig into the `parse` module and figure out why these aren't showing up. They have the X-Curse-ID so even if fingerprinting fails, the backup method should identify them.

**Screenshots**

ajour_YRgIhIAdDf
ajour_6DgGXxaefv
ajour_IP03RYV5gD

tarkah:
d5f9948 resolves #87 

56545d1 resolves #88

### DIFF
--- a/src/fs/addon.rs
+++ b/src/fs/addon.rs
@@ -26,11 +26,19 @@ pub async fn install_addon(
     let mut zip_file = std::fs::File::open(&zip_path)?;
     let mut archive = zip::ZipArchive::new(&mut zip_file)?;
 
-    // TODO: Maybe remove old addon here, so we don't replace.
     for i in 0..archive.len() {
         let mut file = archive.by_index(i)?;
         let path = to_directory.join(file.sanitized_name());
-        if (&*file.name()).ends_with('/') {
+
+        // If top-level destination folder for addon, delete that folder to remove
+        // the previous version so we guarantee a clean copy
+        if let Some(parent) = path.parent() {
+            if parent == to_directory {
+                let _ = std::fs::remove_dir_all(&path);
+            }
+        }
+
+        if file.is_dir() {
             std::fs::create_dir_all(&path)?;
         } else {
             if let Some(p) = path.parent() {
--- a/src/gui/element.rs
+++ b/src/gui/element.rs
@@ -504,21 +504,16 @@ pub fn menu_container<'a>(
     .style(style::DefaultBoxedButton(color_palette));
 
     // Is any addon performing an action.
-    let addons_performing_actions = addons.iter().any(|a| match a.state {
-        AddonState::Downloading | AddonState::Unpacking => true,
-        _ => false,
-    });
+    let addons_performing_actions = addons
+        .iter()
+        .any(|a| matches!(a.state, AddonState::Downloading | AddonState::Unpacking));
 
-    let ajour_performing_actions = match state {
-        AjourState::Loading => true,
-        _ => false,
-    };
+    let ajour_performing_actions = matches!(state, AjourState::Loading);
 
     // Is any addon updtable.
-    let any_addon_updatable = addons.iter().any(|a| match a.state {
-        AddonState::Updatable => true,
-        _ => false,
-    });
+    let any_addon_updatable = addons
+        .iter()
+        .any(|a| matches!(a.state, AddonState::Updatable));
 
     // Enable update_all_button if:
     //   - We have addons.
--- a/src/gui/update.rs
+++ b/src/gui/update.rs
@@ -12,7 +12,7 @@ use {
     iced::{button, Command},
     isahc::HttpClient,
     native_dialog::*,
-    std::path::PathBuf,
+    std::path::{Path, PathBuf},
 };
 
 pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Message>> {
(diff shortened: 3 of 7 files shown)


tarkah:
I gotta run so I can't give a full breakdown on all the optimizations I did, but it was mainly unnecessary allocations (such as cloning just addon found instead of entire vec of addons). 

I've fixed the issues that were preventing addons from parsing after updating. I queue up fingerprinting for the addon as well as any dependencies of the addon. I had to change the `update_addon_fingerprint` function to take the addon name & addon directory vs an `Addon` so I could call it on the dependencies.

src/gui/element.rs
```diff
@@ -504,21 +504,16 @@ pub fn menu_container<'a>(
...
-
-    let ajour_performing_actions = match state {
-        AjourState::Loading => true,
-        _ => false,
-    };
+    let addons_performing_actions = addons
+        .iter()
+        .any(|a| matches!(a.state, AddonState::Downloading | AddonState::Unpacking));
+
+    let ajour_performing_actions = matches!(state, AjourState::Loading);
```
tarkah, line 511: `matches!` makes this much more condensed!

src/gui/update.rs
```diff
@@ -142,28 +142,24 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
         }
         Message::Interaction(Interaction::Delete(id)) => {
             // Delete addon, and it's dependencies.
-            let addons = ajour.addons.clone();
-            if let Some(addon) = addons.iter().find(|a| a.id == id) {
+            if let Some(addon) = ajour.addons.iter().find(|a| a.id == id).cloned() {
```
tarkah, line 145: Instead of cloning the entire vec, just clone the `Option<&Addon>` into `Option<Addon>`

src/gui/update.rs
```diff
@@ -142,28 +142,24 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
...
+                let addons_to_be_deleted = [&addon.dependencies[..], &[addon.id]].concat();
                 let _ = delete_addons(&addon_directory, &addons_to_be_deleted);
-                ajour.addons = ajour
+                ajour
                     .addons
-                    .clone()
-                    .into_iter()
-                    .filter(|a| addons_to_be_deleted.iter().any(|ab| ab != &a.id))
-                    .collect();
+                    .retain(|a| !addons_to_be_deleted.iter().any(|ab| ab != &a.id));
```
tarkah, line 154: I saw you use `retain` in other places. This got rid of another clone :D

src/gui/update.rs
```diff
@@ -330,9 +326,7 @@ async fn perform_download_addon(
 ) -> (String, Result<()>) {
     (
         addon.id.clone(),
-        download_addon(&shared_client, &addon, &to_directory)
-            .await
-            .map(|_| ()),
```
tarkah, line 335: This seemed redundant since it was already `Result<()>`

src/parse.rs
```diff
@@ -98,14 +98,15 @@ pub async fn read_addon_directory<P: AsRef<Path>>(
...
     // If the path does not exists or does not point on a directory we return an Error.
-    if !root_dir.as_ref().is_dir() {
+    if !root_dir.is_dir() {
         return Err(ClientError::Custom(format!(
             "Addon directory not found: {:?}",
-            root_dir.as_ref().to_owned()
+            root_dir
         )));
     }
-    let root_dir = root_dir.as_ref().to_owned();
```
tarkah, line 108: We don't need to make an allocation here, `&Path` worked everywhere it was needed.

src/fs/addon.rs
```diff
@@ -26,11 +26,19 @@ pub async fn install_addon(
...
+
+        // If top-level destination folder for addon, delete that folder to remove
+        // the previous version so we guarantee a clean copy
+        if let Some(parent) = path.parent() {
+            if parent == to_directory {
+                let _ = std::fs::remove_dir_all(&path);
+            }
+        }
+
+        if file.is_dir() {
```
tarkah, line 41: I assume this works over the `if (&*file.name()).ends_with('/') {`?

src/gui/element.rs
casperstorm: > `matches!` makes this much more condensed!

So much cleaner! 🧼

casperstorm:
This has been confirmed to work. I downloaded top 15 addons for retail in older versions and redownloaded them all. After pressing Refresh none showed up as `updatable` which we hoped for. 🙏🏻

Also some really good optimizations so we could trim out a little of the shaky code, allocations, unnecessary clones etc.

PR #91, opened 2020-09-11
parse non-fingerprint curse-id matches

### LINKED ISSUE #90: Addons with unknown fingerprints not shown in UI
**Describe the bug**
Addons that do not match fingerprints known to the curse api are no longer shown in the UI.

**To Reproduce**
1. Pick an addon that is currently shown in the UI (and then close ajour)
2. Edit the addons .toc file with some non-whitespace character. I chose to change the version to some lower number.
3. Delete the fingerprints.yml file to force a fingerprint recalculation on next launch.
4. Open Ajour again and see that the addon is no longer in the list.

Alternatively, use an addon that is not hosted on curseforge in the first place.

**Expected behavior**
Ajour displays all my local addons and uses alternative (legacy/offline) logic to identify local version in case the fingerprint endpoint does not return any matches.

**Screenshots**
01-pre_modification
02-modification
03-post_modification

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Version: 0.3.1

**Additional context**
With the release of 0.3.0 (and 0.3.1) the overall level of support for addons has increased enormously which is great! This issue is only to address that the "fallback" scenario isn't working.

tarkah:
partially addresses #90. We still don't show addons with no API hit.

All addons with a curse id can be queried against curse. Previously, only addons with exact fingerprints got curse package updates. This PR identifies the gap of addons that have a Curse ID, no Tukui ID, and were not successfully fingerprinted. We then apply the curse package.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` and `Removed`.
 
 ## [Unreleased]
+### Fixed
+- Include curse addons that failed fingerprinting
+  - Even though fingerprinting failed, there is still a curse project id that can be used to lookup the addon via the curse API. Previously these addons were not shown, but now they are included.
+
 ## [0.3.2] - 2020-09-11
 ### Changed
 - Light theme is now a bit more gentle to the eyes. Don't worry, it's still not default.
 - Switched Refresh and Update All buttons.
+
 ## [0.3.1] - 2020-09-11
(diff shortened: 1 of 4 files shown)


casperstorm:
Good catch, ill merge this in. @tarkah can you update the CHANGELOG file as well with the changes?

casperstorm (approved):

casperstorm (changes requested):
Good catch and great addition. Ill merge this in. @tarkah can you update the CHANGELOG file as well with the changes?

tarkah:
Just updated, I'll get in the habit of doing that from now on :P

casperstorm (approved):

PR #97, opened 2020-09-12
ensure config dir is created before concurrent operations

### LINKED ISSUE #95: Thread panic at init when run from terminal in Manjaro Linux
**Describe the bug**
Ajour tries to create a folder that already exists and emits a thread panic when run from the terminal under Manjaro KDE.
```sh
thread '<unnamed>' panicked at 'could not create folder $HOME/.config/ajour: Os { code: 17, kind: AlreadyExists, message: "File exists" }', src/fs/mod.rs:24:37
``` 

**To Reproduce**
1. Build ajour
2. Click on the binary in `target/release`
3. Have Dolphin tell you that executing binaries is not allowed for security reasons
4. Run from terminal instead `cd target/release/` -> `./ajour`

Note that this only happens the first time ajour is run in this manner, so I suspect something weird is happening during init. Also, it does not affect functionality, so I wouldn't call it serious. 

**Expected behavior**
No panics!

**Desktop (please complete the following information):**
image

tarkah:
resolves #95

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 ### Fixed
 - Include curse addons that failed fingerprinting
   - Even though fingerprinting failed, there is still a curse project id that can be used to lookup the addon via the curse API. Previously these addons were not shown, but now they are included.
+- Check for & create ajour config folder before launching concurrent init operations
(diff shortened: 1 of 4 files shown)


tarkah:
First thing we do when launching the application is make sure the config directory exists. Once that is done, we can safely batch our init operations (load config, load user themes, check for update).

The issue before was `load_config` and `load_user_themes` were run concurrently and both would try to create the config directory if it didn't exist. When these operations overlapped, it causes a panic since one of them would have already created the directory.

nina-j:
Tested on Manjaro 20.1 Kernel 5.8.6-1 and can confirm that the issue is resolved. :+1:

casperstorm (approved):
Think this is a good approach, since now we have more logic in `update.rs`.
Thanks @tarkah.

PR #98, opened 2020-09-12
check for case insensitive addons folder

### LINKED ISSUE #96: Make AddOns folder search case-insensitive
I'm on Linux. Addon folder was named "Addons" instead of "AddOns", hence Ajour could not find the directory. 

Make folder-search case-insensitive.

Fixed by renaming Addons to AddOns.

**Additional context**
I don't know why my Addon folder was named this way. Probably because I renamed it. I suppose this bug won't happen on Windows.

tarkah:
resolves #96 

The `glob` crate is a little weird. I couldn't get it to match `interface/addons` even with the case insensitive option enabled. I was able to get it working by adding an actual pattern symbol within each component of the path patten, such as `?nterface/?ddons`. With this pattern, I have it working completely case insensitive. I've tested it with a bunch of combinations, including `InTeRFAcE/aDdONs`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 ### Fixed
 - Include curse addons that failed fingerprinting
   - Even though fingerprinting failed, there is still a curse project id that can be used to lookup the addon via the curse API. Previously these addons were not shown, but now they are included.
+- Check for case-insensitive version of `Interface/AddOns` folder for robustness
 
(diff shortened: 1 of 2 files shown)


nina-j:
Tested this while testing #97 because I was already at it - can confirm the issue is solved.

casperstorm:
> Tested this while testing #97 because I was already at it - can confirm the issue is solved.

Thanks for helping us out, @nina-j!

casperstorm (approved):
Great addition. And yes, that glob interface seems a bit strange. Nice you solved it @tarkah.

PR #108, opened 2020-09-13
feat: able to easily select flavor from menu

casperstorm:
This PR makes it easier to switch between flavors (retail/classic). The dropdown menu is moved to the menu row, and both directories are parsed, if there's any addons to parse.

It's a rather big refactor for a small feature, but that's because we are moving away from keeping addons in a `Vec<Addon>` but instead having them in a `HashMap` where `flavor` is key, and `Vec<Addon>` is value. This requires a little bit of plumping here and there.

Feel free to codereview as I go along, @tarkah - it's still a bit messy but you might already have comments.

### DIFF
--- a/src/fs/save.rs
+++ b/src/fs/save.rs
@@ -43,17 +43,8 @@ pub trait PersistentData: DeserializeOwned + Serialize {
         let load_result = <T as PersistentData>::load();
 
         match load_result {
-            Err(ClientError::LoadFileDoesntExist(_)) => Ok(get_default_and_save()?),
-            Err(ClientError::YamlError(error)) => {
-                // Prevent parsing error with an empty string and commented out file.
-                if error.to_string() == "EOF while parsing a value" {
-                    Ok(get_default_and_save()?)
-                } else {
-                    Err(ClientError::YamlError(error))
-                }
-            }
-            Err(error) => Err(error),
             Ok(deser) => Ok(deser),
+            _ => Ok(get_default_and_save()?),
         }
     }
 
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -148,7 +148,7 @@ pub async fn read_addon_directory<P: AsRef<Path>>(
                 .iter()
                 .find(|f| &f.flavor == flavor && &f.title == dir_name)
             {
-                fingerprint.to_owned().clone()
+                fingerprint.to_owned()
             } else {
                 let hash = fingerprint_addon_dir(
                     &addon_dir,
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -104,7 +104,7 @@ pub async fn read_addon_directory<P: AsRef<Path>>(
     fingerprint_collection: Arc<Mutex<Option<FingerprintCollection>>>,
     root_dir: P,
     flavor: Flavor,
-) -> Result<(Flavor, Vec<Addon>)> {
+) -> Result<Vec<Addon>> {
     let root_dir = root_dir.as_ref();
 
     // If the path does not exists or does not point on a directory we return an Error.
@@ -317,7 +317,7 @@ pub async fn read_addon_directory<P: AsRef<Path>>(
 
     // Concats the different repo addons, and returns.
     let concatenated = [&fingerprint_addons[..], &tukui_addons[..]].concat();
-    Ok((flavor, concatenated))
+    Ok(concatenated)
 }
 
 pub async fn update_addon_fingerprint(
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -28,6 +28,7 @@ lazy_static::lazy_static! {
 pub struct Fingerprint {
     pub title: String,
     pub hash: Option<u32>,
+    pub flavor: Flavor,
 }
 
 pub type FingerprintCollection = Vec<Fingerprint>;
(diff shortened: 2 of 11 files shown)


src/fs/save.rs
```diff
@@ -43,17 +43,8 @@ pub trait PersistentData: DeserializeOwned + Serialize {
...
-                // Prevent parsing error with an empty string and commented out file.
-                if error.to_string() == "EOF while parsing a value" {
-                    Ok(get_default_and_save()?)
-                } else {
-                    Err(ClientError::YamlError(error))
-                }
-            }
-            Err(error) => Err(error),
             Ok(deser) => Ok(deser),
+            _ => Ok(get_default_and_save()?),
```
casperstorm, line 47: If anything goes wrong loading fingerprint.yml I really just want us to fallback to default rather than throwing error.

src/parse.rs
```diff
@@ -156,22 +159,36 @@ pub async fn read_addon_directory<P: AsRef<Path>>(
...
+        .clone()
+        .into_iter()
+        .filter(|f| &f.flavor != flavor)
+        .collect();
+
     // Update our in memory collection and save to disk.
     fingerprint_collection.drain(..);
-    fingerprint_collection.extend(new_fingerprints);
+    fingerprint_collection
+        .extend([&new_fingerprints[..], &other_flavors_fingerprints[..]].concat());
```
casperstorm, line 180: @tarkah – This is a place I could use your help for optimizing.
What is happening is that i've extended our type we save to `fingerprints.yml` with flavor. Before, everytime we switched flavor, we removed the fingerprints. Now we have them all, for both flavors.
That means I have to do some filtering on our fingerprint_collection when using it, and when saving `new_fingerprints` i need to save "the others" as well, otherwise we would only have either `retail` or `classic` – however it's not that pretty the solution ive made right now so if you have some idea up your sleeve, please let me know or push a fix.

tarkah: We need to change fingerprint collection into a hashmap with flavor as key as well. This way we can just replace the Vec<Fingerprint> value for that flavor key and save the collection. No need to worry about the other flavor since that side won't get modified.

src/fs/save.rs
tarkah: I think this is fine for our persistent data trait since nothing should be touched by the end user. If we had a user controlled config file, this would be dangerous to do since we could easily overwrite their changes.

src/parse.rs
casperstorm: Maybe it's something you can help with? We need to think about mirgation, so the current users then so they don't have to do anything.

tarkah: Yeah I can help. With your change to the Persistent Data trait, it won't fail. New fingerprints will just get generated on first launch, which shouldn't be a big deal for a new release.

casperstorm: @tarkah you are the best. I won't touch anything in `parse.rs` until you have pushed then to avoid a nasty merge.
I'll focus on everything around and make sure that is working and plays nicely.

tarkah: See #109

tarkah:
So every time I launch or refresh, I am getting the error message for retail that it can't find the addons directory. We should suppress this error unless it's generated for the selected flavor (I don't want to see it if I have Classic selected, just because its also trying to parse Retail)

casperstorm:
> So every time I launch or refresh, I am getting the error message for retail that it can't find the addons directory. We should suppress this error unless it's generated for the selected flavor (I don't want to see it if I have Classic selected, just because its also trying to parse Retail)
> 
> ajour_uIM0hqXyyO

Yes! Thanks, @tarkah.
I havnt done this logic just yet so I am aware. Ill work on it over the next few hours.

casperstorm:
This is ready for review and testing.
I'll properly make a RC of this tomorrow when I wake up.

tarkah (changes requested):
Just finished reviewing and testing, looks good!

There is one issue, however. Ignored addons are not flavor aware. If I ignore a classic addon, it also shows up in the settings->Ignore Addons list if I switch to retail. Also, it doesn't appear ignored addons are getting saved to the config file, so it doesn't persist across sessions.

casperstorm:
@tarkah ignored_addons are now flavor aware. One issue is that it will default our config on boot, because of the new format - meaning wow path is lost. I am trying to see if I can do some migration because it's not so fun to lose that.

tarkah:
@casperstorm I just saw your latest push. This is a breaking change now to the config, right? We need to make sure when Config gets deserialized that it doesn't fail on this and default the entire config. We should make a custom deserializer that first tries to deser it as Vec and then applies that vec to both flavors so they don't lose. Or create an empty hashmap for just the ignored add-ons.

tarkah:
Lol I just commented the same :P

casperstorm:
> Lol I just commented the same :P

I'll try to work on a migration on it now :)

tarkah:
Awesome! We can define a deserialize_with function I think

casperstorm:
> deserialize_with

Yeah like you did with `theme`.

tarkah:
Yup!

tarkah:
Check out #112 for the deser logic on ignored addons!

PR #109, opened 2020-09-14
change fingerprint collection to hashmap

tarkah:
for #108

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -28,10 +28,16 @@ lazy_static::lazy_static! {
 pub struct Fingerprint {
     pub title: String,
     pub hash: Option<u32>,
-    pub flavor: Flavor,
 }
 
-pub type FingerprintCollection = Vec<Fingerprint>;
+#[derive(Serialize, Deserialize, Default)]
+pub struct FingerprintCollection(HashMap<Flavor, Vec<Fingerprint>>);
+
+impl FingerprintCollection {
+    fn get_mut_for_flavor(&mut self, flavor: Flavor) -> &mut Vec<Fingerprint> {
+        self.0.entry(flavor).or_default()
+    }
+}
 
 impl PersistentData for FingerprintCollection {
     fn relative_path() -> PathBuf {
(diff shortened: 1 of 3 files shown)


src/parse.rs
```diff
@@ -97,7 +103,7 @@ async fn file_parsing_regex() -> Result<ParsingPatterns> {
 pub async fn read_addon_directory<P: AsRef<Path>>(
     fingerprint_collection: Arc<Mutex<Option<FingerprintCollection>>>,
     root_dir: P,
-    flavor: &Flavor,
+    flavor: Flavor,
```
tarkah, line 106: Things that are `Copy` should just get passed as owned, not a reference

src/parse.rs
```diff
@@ -137,17 +143,15 @@ pub async fn read_addon_directory<P: AsRef<Path>>(
...
+    let new_fingerprints: Vec<_> = all_dirs
         .par_iter() // Easy parallelization
         .map(|dir_name| {
             let addon_dir = root_dir.join(dir_name);
             // If we have a stored fingerprint on disk, we use that.
-            if let Some(fingerprint) = fingerprint_collection
-                .iter()
-                .find(|f| &f.flavor == flavor && &f.title == dir_name)
-            {
+            if let Some(fingerprint) = fingerprints.iter().find(|f| &f.title == dir_name) {
```
tarkah, line 154: We no longer need to filter on flavor since this data set is now specific for the flavor

src/parse.rs
```diff
@@ -159,36 +163,22 @@ pub async fn read_addon_directory<P: AsRef<Path>>(
...
         .filter(|f| f.hash.is_some())
         .collect();
 
-    // Get the other flavors, we want saved as well.
-    let other_flavors_fingerprints: Vec<_> = fingerprint_collection
-        .clone()
-        .into_iter()
-        .filter(|f| &f.flavor != flavor)
-        .collect();
-
```
tarkah, line 176: Yay, we can get rid of this now!

casperstorm: **AWESOME**.

casperstorm:
This is great. Its tested and is working flawless as always.

PR #112, opened 2020-09-15
deser old ignored addons format

tarkah:
@casperstorm I got it figured out! Check out the `visit_seq` for the meat of how we handle the old array format. I tested this and it works!

I also noticed that ingored addons in the settings only showed up for the flavor at launch, but if I switched to the other flavor, it didn't show up (unless you hit refresh). I fixed in the ParsedAddons message logic so both flavors have their ignored addons added.

### DIFF
--- a/src/gui/update.rs
+++ b/src/gui/update.rs
@@ -235,17 +235,17 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
                 if flavor == ajour.config.wow.flavor {
                     // Set the state if flavor matches.
                     ajour.state = AjourState::Idle;
-
-                    // Find and push the ignored addons.
-                    let ignored_ids = ajour.config.addons.ignored.entry(flavor).or_default();
-                    let ignored_addons: Vec<_> = addons
-                        .iter()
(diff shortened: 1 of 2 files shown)


src/gui/update.rs
```diff
@@ -235,17 +235,17 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
...
+                let ignored_ids = ajour.config.addons.ignored.entry(flavor).or_default();
+                let ignored_addons: Vec<_> = addons
+                    .iter()
+                    .filter(|a| ignored_ids.iter().any(|i| i == &a.id))
+                    .map(|a| (a.clone(), button::State::new()))
+                    .collect::<Vec<(Addon, button::State)>>();
 
-                    ajour.ignored_addons.insert(flavor, ignored_addons);
-                }
+                ajour.ignored_addons.insert(flavor, ignored_addons);
```
tarkah, line 248: This should happen on the flavor we don't have selected after parsing as well :)

casperstorm:
> @casperstorm I got it figured out! Check out the `visit_seq` for the meat of how we handle the old array format. I tested this and it works!

Ah, yes – that make sense actually. I saw you have different methods for the Visitor. Thats smart.

Issue #101, opened 2020-09-13
Make Columns draggable and resizable


Ruffz:
_**Is your feature request related to a problem? Please describe.**_
I would like to have "Status" column right after "Addon" column and have ability to set the space between columns.

_**Describe the solution you'd like**_
Make Columns draggable and resizable.

casperstorm:
Thank you so much for your feedback, @Ruffz.
I have no added it to our roadmap: https://github.com/casperstorm/ajour/projects/2#card-45550101.

Ruffz:
Thanks for this update²

Columns resize is perfect, but I can't drag/reorder these

casperstorm:
> Thanks for this update²
> 
> 
> 
> Columns resize is perfect, but I can't drag/reorder these

Great!
Reorder will come later. It requires quite a refactor.

PR #119, opened 2020-09-17
header widget

tarkah:
@casperstorm This isn't in any state that we'd want to use, but I wanted to proof of concept out what using the `PaneGrid` would look like for resizable columns.

Try it out, it actually works! However, it's SUPER hacky and has a LOT of limitations and things that just don't work the way they should. For example, since you can't actually specify the width of the panes, just the ratio of the split between two panes, I basically had to hard code what those split ratios look like to get the default column size we currently have. And that only works for the default program width, it completely breaks when resizing the application.

I'm convinced through this test that we just need to build our own table widget to support resizing, reordering, adding / removing columns, etc. I had a good read through the `pane_grid` module and I found your `Iced` issue that hercj responded to with his ideas around the API and I think it's totally doable. If I do create a custom widget for this, I'll probably break it out into it's own crate and if it works well, I can submit it as a PR to `Iced`.

Should any crates / libs we create go under a `crates` folder in the root of this project?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 - It is now possible to see when a update was released in the details view of an addon. 
 - A website button has been added to the detail view of each addon.
   - The idea is that with this button, it is easy to open up the addon website and view additional information which Ajour might not show.
+- Columns can be resized by clicking & dragging the dividers between the column headers. This change will be saved and used when starting Ajour.
+- Window size will be saved when resizing the application and used when starting Ajour.
 
 ### Changed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ edition = "2018"
 [dependencies]
 iced = { git = "https://github.com/hecrj/iced.git", features = ["debug"], rev = "fb015a85d22a7c4632bd251127a89259bfd0c346" }
 iced_futures = { git = "https://github.com/hecrj/iced.git", features = ["async-std"], rev = "fb015a85d22a7c4632bd251127a89259bfd0c346" }
+iced_native = { git = "https://github.com/hecrj/iced.git", rev = "fb015a85d22a7c4632bd251127a89259bfd0c346" }
 regex = "1.3.9"
 fancy-regex = "0.3.5" # Regex with backtracking
 async-std = "1.6.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,5 +36,13 @@ log = "0.4"
 fern = "0.6"
 timeago = "0.2.1"
 
+widgets = { path = "crates/widgets" }
+
 [build-dependencies]
 embed-resource = "1.3.3"
+
+[workspace]
+members = [
+    ".",
+    "crates/widgets"
+]
--- a/crates/widgets/src/renderer/table_header.rs
+++ b/crates/widgets/src/renderer/table_header.rs
@@ -1,0 +1,44 @@
+use crate::widget::table_header;
+use iced_graphics::{Backend, Primitive, Renderer};
+use iced_native::mouse;
+use iced_native::{Element, Layout, Point};
+
+impl<B> table_header::Renderer for Renderer<B>
+where
+    B: Backend,
+{
+    fn draw<Message>(
+        &mut self,
+        defaults: &Self::Defaults,
+        content: &[Element<'_, Message, Self>],
+        layout: Layout<'_>,
+        cursor_position: Point,
+        resize_hovering: bool,
+    ) -> Self::Output {
+        let mut mouse_interaction = if resize_hovering {
+            mouse::Interaction::ResizingHorizontally
+        } else {
+            mouse::Interaction::default()
+        };
+
+        (
+            Primitive::Group {
+                primitives: content
+                    .iter()
+                    .zip(layout.children())
+                    .map(|(child, layout)| {
+                        let (primitive, new_mouse_interaction) =
+                            child.draw(self, defaults, layout, cursor_position);
+
+                        if new_mouse_interaction > mouse_interaction {
+                            mouse_interaction = new_mouse_interaction;
+                        }
+
+                        primitive
+                    })
+                    .collect(),
+            },
+            mouse_interaction,
+        )
+    }
+}
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -24,6 +24,8 @@ pub struct Config {
 
     #[serde(default)]
     pub column_config: ColumnConfig,
+
+    pub window_size: Option<(u32, u32)>,
 }
 
 impl Config {
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -21,6 +21,9 @@ pub struct Config {
     pub addons: Addons,
 
     pub theme: Option<String>,
+
+    #[serde(default)]
+    pub column_config: ColumnConfig,
 }
 
 impl Config {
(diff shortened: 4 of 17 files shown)


casperstorm:
Ha, @tarkah! This actually works kinda nice.
I agree it is sorta hacky, but on the other hand - what isn't in Ajour now that we use iced? I mean, we do jump some hoops to bend the UI the way we want to archive good UX. What brothers me right now is that you **expect** to be able to resize because its a common thing, but you can't so how do we get there? This could most certainly be a good way.

One thing I was talking to @mlablah about earlier today is that whenever you resize something, we should properly save the value for the next session.

tarkah:
Yeah I was surprised how well it actually "just worked". The thing that bothers me about how it works though is that because the panes are "nested", the resizing is wonky for anything that isn't the 1st / 2nd pane / column (leftmost in how I set it up). Try resizing the status column and see what I mean. There is no way around this due to the nested layout nature of the panes / splits.  So I definitely don't recommend using this hack to ship this feature.

But considering the resizing works REALLY nicely/ smoothly, I can pretty much just copy that logic from the `pane_grid` module of Iced and add it to a custom Table widget. Which I think is the hardest part about it. Everything else should be _fairly_ easy to implement.

casperstorm:
Good point. That would actually be nice next step. I also think it's important to remember the state we are in and what this would add. It is a enhancement feature. What I mean is that I agree with you that we shouldn't jump too soon and "bloat" the code base with a hack just to get resizing. I mean its a REALLY nice feature, and I would love to have it - but for what cost right. This work is really nice because it shows us that it is indeed possible to do #101 with iced.

tarkah:
Exactly! I've started on something that I think may work... it's basically a glorified wrapper around `Scrollable` that takes some addition state for tracking column widths on resize events. It also lets you add an entire "Column (header + row data)" definition. When you add the "Column", you pass it an Iterator of the data, a closure which maps that data to an Element (what will display on the row), and a closure that returns an Element for the Header. Then I basically build a Column from those components and apply any saved column width from the state.

Here's how that roughly looks:

```rust
    pub fn add_column<D>(
        mut self,
        column_name: &'a str,
        initial_width: Length,
        data: impl Iterator<Item = &'a D>,
        header_view: impl Fn(&str) -> Element<'a, Message, Renderer>,
        row_view: impl Fn(&'a D) -> Element<'a, Message, Renderer>,
    ) -> Self
    where
        D: 'a,
    {
        let width = self
            .state
            .column_width(&column_name)
            .unwrap_or(initial_width);

        let header = header_view(&column_name);

        let mut column = Column::new().width(width).push(header);

        for elem in data.map(move |data| row_view(data)) {
            column = column.push(elem);
        }

        self.content = self.content.push(column);

        self
    }
```

I need to test this out and make sure it displays properly. The only caveat with this is you need to pass in the column definitions in the order you want them displayed vs having order retained inside the State. This is a limitation since I'm just pushing everything inside a `Scrollable`. But I only plan on initially adding in resize logic, not draggable / reordering, so I think this is fine. That can be a phase 2 thing.

I'm going to be out of town this weekend so I probably won't make any further progress on this until next week.

tarkah:
@casperstorm I need to go through and clean some things up and add config persistence, but I've got it in a working state! Try it out!

I basically created a wrapper around `Row` that allows me to track when the cursor is around the divider between row elements and calculate a new width for the element to the left / right of that divider. It will send a ResizeEvent message when this occurs which I've then handled inside our gui module.

crates/widgets/src/renderer/table_header.rs
```diff
@@ -0,0 +1,44 @@
...
+        content: &[Element<'_, Message, Self>],
+        layout: Layout<'_>,
+        cursor_position: Point,
+        resize_hovering: bool,
+    ) -> Self::Output {
+        let mut mouse_interaction = if resize_hovering {
+            mouse::Interaction::ResizingHorizontally
+        } else {
+            mouse::Interaction::default()
+        };
```
tarkah, line 22: This makes our cursor change when hovering over a divider between row elements

tarkah, line 22: This makes our cursor change when hovering over a divider between row elements

src/config/mod.rs
```diff
@@ -87,6 +90,42 @@ impl PersistentData for Config {
...
 }
 
+#[derive(Deserialize, Serialize, Debug, PartialEq, Clone)]
+pub enum ColumnConfig {
+    V1 {
+        local_version_width: u16,
+        remote_version_width: u16,
+        status_width: u16,
+    },
+}
```
tarkah, line 100: I've added this as an enum so we don't break deserialization in the future if this needs to change (such as adding ordering & hidden).

tarkah, line 100: I've added this as an enum so we don't break deserialization in the future if this needs to change (such as adding ordering & hidden).

tarkah: Also, "title" width isn't defined here since I always leave "title" as `Length::Fill". This ensures the table still resizes nicely when the application does.

tarkah:
I've also just added window resizing persistence. I feel like this needs to roll out with the column resizing option in case someone resizes their columns within a resized window to keep everything correct.

Because I am now accessing the `Config` within `gui::run`, which occurs before any of the async commands, I've reverted the fix we implemented back from #97 since that's no longer at risk.

tarkah:
Should we add a button in Settings such as `reset layout` or the like, in case the user wants to revert window sizing / column sizing back to the default? Otherwise they'd have to open the config file and delete those lines out.

tarkah:
Also, I rebased this off #123 but I can rebase to master if you want to try and merge this first

CHANGELOG.md
```diff
@@ -13,6 +13,15 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 
 - Windows: Ajour now comes bundled with the needed dependencies. This way we avoid relying on the system having Microsoft Visual C++ 2015 Redistributable.
 
+### Added
+
+- A website button has been added to the detail view of each addon.
+  - The idea is that with this button, it is easy to open up the addon website and view additional information which Ajour might not show.
+
```
casperstorm, line 20: Can you add an entry to the CHANGELOG.md?

Cargo.toml
```diff
@@ -35,5 +36,13 @@ chrono = "0.4"
 log = "0.4"
 fern = "0.6"
 
+widgets = { path = "crates/widgets" }
```
casperstorm, line 39: I think it's fine to have it in `crates/widgets` - maybe too specific to break out of ajour, and too generic to stay inside the `src` folder. I think this is a good solution.

src/config/mod.rs
```diff
@@ -21,6 +21,11 @@ pub struct Config {
     pub addons: Addons,
 
     pub theme: Option<String>,
+
+    #[serde(default)]
+    pub column_config: ColumnConfig,
+
+    pub window_size: Option<(u32, u32)>,
```
casperstorm, line 28: Great addition we actually save it to config. This is were it becomes really useful.

src/config/mod.rs
casperstorm: Make sense, good thinking.

casperstorm (approved):
This is a really useful addition to Ajour. It does add some complexity now that we have a custom `Iced` widget, but I do believe it's necessary for the UX. Good job, @tarkah. This was a hard one.

casperstorm:
Another thing to consider: how often are we writing to the log when resizing? I am just wondering about if we should only write during mouse lift, I dunno if this could be a problem with all these writes.

tarkah:
I've updated the changelog and just completely removed the logging for resize events. It's all noise and isn't something we need polluting the log, on top of all the writes it'd cause. I looked into `fern` to see if they use some kind of buffering / flushing to the log file, but couldn't find any documentation on it, so I think it's best to assume worst case scenario there.

casperstorm (approved):
It looks good man, really good.
Do you have any more you want to add?

tarkah:
Thanks! I think I'm happy with where it's at. We can totally add some more customization down the road, to support reordering and a overlay menu to select / deselect columns for visibility. But that might take me some time to get around to.

PR #122, opened 2020-09-19
feat: add png icons

masonnl:
_disclaimer: I'm very new to contributing to open source projects. Any feedback regarding my contribution is welcome and appreciated!_

## Context for changes
As a Linux user, I would like to create desktop shortcuts to the ajour app and use the ajour logo as the shortcut icon.

## Proposed Changes
  - Uses the [`windows/ajour.ico`](https://github.com/casperstorm/ajour/blob/master/resources/windows/ajour.ico) and Gimp to export png images of the ajour icon in various sizes.
- Place the new pngs into a `resources/logo/` directory
  - 16x16, 24x24, 32x32, 48x48, 256x256, 1024x1024

_48x48 example:_
ajour_icons_on_linux

## Checklist

- [n/a] Tested on all platforms changed
- [n/a] `cargo fmt` has been run on this branch
- [n/a] `cargo clippy` has been run on this branch
- [n/a] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users


masonnl:
Since I do not have access to an SVG, these image sizes are ripped directly from the ico. This is probably fine for now.

Maybe these PNGs should live somewhere else though? It doesn't seem like the right place to save asset media in a linux directory. I'd like to hear @casperstorm's thoughts on where the PNGs might best belong if you even want them in the first place! Thanks!

casperstorm:
Hi @masonnl.
Thank you for contributing to Ajour!

This is an nice addition to Linux users for sure, however I am a little unsure - you manually created the shortcut and then added icons from the folder? 
In any case I don't think we should put generic logos into a platform specific folder. I would rather have them living in `resources/logo/`. What do you say?

masonnl:
> you manually created the shortcut

Yes. Once the binary was built to `/target/release/ajour` I wanted a way to run it from my taskbar so I created the shortcut.

> and then added icons from the folder?

Yes. Once the shortcut was made, I tried changing the icon of the shortcut to the .ico under `windows/` but discovered the linux doesn't play well with .ico files well out of the box. The shortcut appeared blurry and was not automatically adapting to one of the icon's pre-defined dimensions.

My solution to this was to use Gimp to open the .ico and extract each layer to a separate file and then export the different sizes as standalone pngs. But looking at the diff here it seems Github is reading them as binary files for some reason. Maybe whatever export I used was not ideal for png files. I'll take a look at that.

Thanks for the feedback! I can move them into `logo/`

masonnl:
RE: GH reading the pngs as binary

Looks like this is normal. On the diff, selecting "Display the rich diff" will read the file as png and display the image as expected. Just caught me off guard

casperstorm:
This looks good.
Is there more you have in your mind since the PR is in draft mode, @masonnl?

masonnl:
That was it! Thanks

PR #124, opened 2020-09-23
add scaling

tarkah:
Resolves https://github.com/casperstorm/ajour/projects/2#card-45378181

This adds the ability to specify application scale in the settings & persists that change to the config. Overall an easy change since `Iced` provides a scale factor on the application so we don't have to manually scale anything.

### DIFF
--- a/src/gui/element.rs
+++ b/src/gui/element.rs
@@ -94,7 +94,7 @@ pub fn settings_container<'a>(
 
     // Scale buttons for application scale factoring.
     let (scale_title_row, scale_buttons_row) = {
-        let scale_title = Text::new("Scale").size(DEFAULT_FONT_SIZE);
+        let scale_title = Text::new("UI Scale").size(DEFAULT_FONT_SIZE);
         let scale_title_row = Row::new().push(scale_title).padding(DEFAULT_PADDING);
 
         let scale_down_button: Element<Interaction> = Button::new(
--- a/src/gui/element.rs
+++ b/src/gui/element.rs
@@ -1,7 +1,7 @@
 use {
     super::{
         style, Addon, AddonState, AjourState, ColorPalette, Config, Flavor, HeaderState,
-        Interaction, Message, SortDirection, SortKey, ThemeState,
+        Interaction, Message, ScaleState, SortDirection, SortKey, ThemeState,
     },
     crate::VERSION,
     chrono::prelude::*,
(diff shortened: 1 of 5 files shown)


src/gui/element.rs
```diff
@@ -172,7 +217,9 @@ pub fn settings_container<'a>(
         .push(right_spacer);
 
     // Returns the final container.
-    Container::new(row).height(Length::Units(130))
+    Container::new(row)
+        .height(Length::Shrink)
+        .style(style::AddonRowDefaultTextContainer(color_palette))
```
tarkah, line 222: I changed the entire container to Shrink and then gave the right column a fixed height. This renders the settings page better when the width is reduced and the text starts wrapping, making the left column extend. Previously, if you resized the width down, the left column "Theme" button would extend the settings container and overlay the addons table.

src/gui/update.rs
```diff
@@ -544,6 +575,9 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
         Message::RuntimeEvent(iced_native::Event::Window(
             iced_native::window::Event::Resized { width, height },
         )) => {
+            let width = (width as f64 * ajour.scale_state.scale) as u32;
+            let height = (height as f64 * ajour.scale_state.scale) as u32;
+
```
tarkah, line 580: This tripped me up so much. Apparently once you apply scaling, the Resized event returns a "scaled" width and height. So I had to adjust it back before saving. I was getting tripped up why the window would open at different sizes anytime I changed the scale.

src/gui/element.rs
```diff
@@ -91,6 +92,48 @@ pub fn settings_container<'a>(
     // Data row for theme picker list.
     let theme_data_row = Row::new().push(left_spacer).push(theme_pick_list);
 
+    // Scale buttons for application scale factoring.
+    let (scale_title_row, scale_buttons_row) = {
+        let scale_title = Text::new("Scale").size(DEFAULT_FONT_SIZE);
```
casperstorm, line 97: Should we write `UI Scale` here instead? Just to make it explicit what you are scaling.

casperstorm (changes requested):

casperstorm (changes requested):
Can you also update CHANGELOG.md, @tarkah?

tarkah:
I still always forget about the CHANGELOG :P Agreed on the title, I updated it to UI Scale. This should be good to go!

casperstorm:
> I still always forget about the CHANGELOG :P Agreed on the title, I updated it to UI Scale. This should be good to go!

Awesome 😎👏
I'll merge this in a few hours.

PR #125, opened 2020-09-23
backup feature

tarkah:
This feature will allow the user to backup their Addons and WTF folder through Ajour. Once the user selects a backup directory in settings, a "Backup Now" button will be available to push. Once clicked, the app will zip the Addons and WTF folders from each flavor present on the system and save that zip file under the backup folder with name `ajour_backup_YYYY-MM-DD_HH-MM-SS.zip`. We can then parse this folder on startup and determine the most recent backup made by the user and present that datetime along with the Backup Now button.

This won't manage the backups once they are created, so it won't handle pruning old backups or restoring from backups.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 - Columns can be resized by clicking & dragging the dividers between the column headers. This change will be saved and used when starting Ajour.
 - Window size will be saved when resizing the application and used when starting Ajour.
 - UI Scaling has been added to settings. UI scale can be increased or decreased and will be saved when changed.
+- A backup option has been added to archive the AddOns and WTF folders from each flavor installed on the machine to a user chosen directory.
+  - Backups created through Ajour are not actively managed once created, so pruning old backups and restoring from a backup need to be handled by the user
 
 ### Changed
 
--- a/src/backup.rs
+++ b/src/backup.rs
@@ -1,0 +1,90 @@
+use crate::error::ClientError;
+use crate::fs::backup::{Backup, ZipBackup};
+use crate::Result;
+
+use chrono::{Local, NaiveDateTime};
+use std::convert::TryFrom;
+use std::path::{Path, PathBuf};
+
+/// Creates a .zip archive from the list of source folders and
+/// saves it to the dest folder.
+pub async fn backup_folders(
+    src_folders: Vec<BackupFolder>,
+    mut dest: PathBuf,
+) -> Result<NaiveDateTime> {
+    let now = Local::now();
+
+    dest.push(format!(
+        "ajour_backup_{}.zip",
+        now.format("%Y-%m-%d_%H-%M-%S")
+    ));
+
+    let zip_backup = ZipBackup::new(src_folders, &dest);
+
+    zip_backup.backup()?;
+
+    let as_of = Archive::try_from(dest)?.as_of;
+
+    Ok(as_of)
+}
+
+/// Finds the latest archive in the supplied backup folder and returns
+/// the datetime it was saved
+pub async fn lastest_backup(backup_dir: PathBuf) -> Option<NaiveDateTime> {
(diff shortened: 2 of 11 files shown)


tarkah:
@casperstorm Let me know if you agree with my outline of this PR above. My first commit handles the logic for "backing" up folders with zip format. I've implemented a test for this backup, but we will probably want to remove that before this PR goes live since it operates on the actual fs.

casperstorm:
> Let me know if you agree with my outline of this PR above. My first commit handles the logic for "backing" up folders with zip format. I've implemented a test for this backup, but we will probably want to remove that before this PR goes live since it operates on the actual fs.

Looks good to me, @tarkah. Exactly how I thought of it as well. Looks nice and clean as well.
I agree with removing the test once we are ready to merge this.

tarkah:
All done! I think it works pretty well. The only change I'd like to make is to have some sort of "animation" that the backup process is in progress, vs the static text I currently show. I haven't really looked into how that can be done with Iced.

Also, I expanded the Settings height to fit the backup options in the right column, and leave enough space to show 4 ignored addons within the scrollable. Hopefully we don't have too much more to squeeze into the settings pane :P

tarkah:
Some screenshots to show the change from No Directory Selected -> Selected -> Backing Up -> Backup Done

image
image
image
image

src/backup.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    let as_of = Archive::try_from(dest)?.as_of;
+
+    Ok(as_of)
+}
+
+/// Finds the latest archive in the supplied backup folder and returns
+/// the datetime it was saved
+pub async fn lastest_backup(backup_dir: PathBuf) -> Option<NaiveDateTime> {
+    let pattern = format!("{}/ajour_backup_[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]_[0-9][0-9]-[0-9][0-9]-[0-9][0-9].zip", &backup_dir.display());
```
casperstorm, line 34: I like that we find the latest saved rather than saving it to the config. Good thinking here @tarkah.

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 - Columns can be resized by clicking & dragging the dividers between the column headers. This change will be saved and used when starting Ajour.
 - Window size will be saved when resizing the application and used when starting Ajour.
 - UI Scaling has been added to settings. UI scale can be increased or decreased and will be saved when changed.
+- A backup option has been added to archive the AddOns and WTF folders from each flavor installed on the machine to a user chosen directory.
+  - Backups created through Ajour are not actively managed once created, so pruning old backups and restoring from a backup need to be handled by the user
```
casperstorm, line 25: You remembered it 😀

src/gui/mod.rs
```diff
@@ -41,7 +42,7 @@ pub enum Interaction {
     Delete(String),
     Expand(String),
     Ignore(String),
-    OpenDirectory,
+    OpenDirectory(DirectoryType),
```
casperstorm, line 45: 👍🏻

casperstorm (approved):
This is tested on both Windows and macOS.
On Windows I had no issues at all, but on macOS it seemed like there was a problem unzipping. I was using the default `Archive Utility.app` – i then tried `The Unarchiver.app` a popular third party and that worked. I also used the default `unzip` command from the terminal and that worked as well so I believe it must be (yet another) bug in either macOS or `Archive Utility.app` – I don't want to hold this feature back because of it.

Good job, @tarkah – really a nice solution you came up with.

tarkah:
Interesting, you'd think the default app would work. I'm using the bzip2 compression method, maybe it doesn't support that? Anyway, I also tried on linux and confirmed it worked with the gnome archive manager.

casperstorm:
> Interesting, you'd think the default app would work. I'm using the bzip2 compression method, maybe it doesn't support that? Anyway, I also tried on linux and confirmed it worked with the gnome archive manager.

https://github.com/casperstorm/ajour/pull/132

PR #126, opened 2020-09-23
fix resizing bounds

tarkah:
I noticed a bug in the column resizing that I somehow didn't notice before. My intent was to only have columns be resizable when hovering over the actual header. However I noticed the column could be resized anywhere along the y axis, as long as your cursor was over the correct x axis spot. I've fixed the widget to only allow resizing / change of mouse icon when hovering over the divider while inside the header bounds.

### DIFF
--- a/crates/widgets/src/widget/header.rs
+++ b/crates/widgets/src/widget/header.rs
@@ -150,88 +150,100 @@ where
         renderer: &Renderer,
         clipboard: Option<&dyn Clipboard>,
     ) {
-        let child_len = self.children.len();
-        let start_offset = if self.left_margin { 1 } else { 0 };
-        let end_offset = if self.right_margin { 1 } else { 0 };
+        let in_bounds = layout.bounds().contains(cursor_position);
 
-        let dividers = self
-            .children
-            .iter()
-            .enumerate()
-            .zip(layout.children())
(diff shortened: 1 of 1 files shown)


crates/widgets/src/widget/header.rs
```diff
@@ -150,88 +150,100 @@ where
...
-            .enumerate()
-            .zip(layout.children())
-            .filter_map(|((idx, _), layout)| {
-                if idx >= start_offset && idx < (child_len - 1 - end_offset) {
-                    Some((idx, layout.position().x + layout.bounds().width))
-                } else {
-                    None
+        let in_bounds = layout.bounds().contains(cursor_position);
+
+        if self.state.resizing || in_bounds {
```
tarkah, line 155: This is the only relevant part of the diff, everything else is just from getting moved into the if statement.

crates/widgets/src/widget/header.rs
```diff
@@ -150,88 +150,100 @@ where
...
+                            messages,
+                        );
+                        return;
+                    }
                 }
+                _ => {}
             }
-            _ => {}
+        } else {
+            self.state.resize_hovering = false;
```
tarkah, line 246: and this part

casperstorm:
Ah good catch. Don't know why I didn't notice this 🤔

PR #129, opened 2020-09-24
feat/release channels

### LINKED ISSUE #92: Able to select Release Type of Alpha or Beta when updating addons
I prefer to install Alpha releases of addons (and as I have 185 addons installed it's not really feasible to do it individually). When an alpha version is released I install it, and then only revert to Beta or Release if there are issues found.

Ability to select the release type of individual addons (e.g. DBM as alpha release, KuiNameplates as Release, Raven as Alpha)

N/A

casperstorm:
Resolves #92 

## Proposed Changes
This feature will refactor Ajour into having support for `beta`, `alpha` and `stable` release channels.

## Checklist

- [x] `cargo fmt` has been run on this branch
- [x] `cargo clippy` has been run on this branch


casperstorm:
@tarkah This is mostly done now.
I am still missing to run it though proper testing and clippy, but all the logic should be in place and ready for testing if you have the time. 

Idea is that we have removed all the `remote_*` from `Addon`. Instead we have a `release_channel` and `remote_packages`. A `release_channel` can be either `Stable`, `Beta`,  or `Alpha`. `remote_packages` is a Hashmap, where key is a `release_channel` and value is:

```
pub struct RemotePackage {
    pub version: String,
    pub download_url: String,
    pub date_time: DateTime<Utc>,
    pub is_update: bool,
}
```

This way each addon can hold information about all `release_channels` at once, and we can switch between. Take it for a spin if you have the time. I'll close it up tomorrow.

tarkah:
Awesome! I'll take a look here shortly.

tarkah:
A couple immediate thoughts...

- If you switch to Alpha, it shows an "Update" button. However, if you update and want to switch back to Stable, no button is available to switch.
- Should we show release channels in the pick list if they aren't available for the addon?
- I noticed a lot of the time, the Alpha or Beta channel are far behind Stable. I'd imagine most people are only interested in those channels if they are newer than the latest stable release? It'd be nice to know when the Alpha / Beta channel for an addon is "outdated" to stable if currently selected.
- Likewise, it might be good to know when a new Alpha / Beta release is available compared to the installed Stable addon. 
- I guess for the previous two points, its actually hard to make these determinations since release date isn't helpful if the older Beta or Alpha release are still a "newer" more bleeding edge version being tested, so maybe we can't really do much here? It's not like everyone is using semantic versioning :P 
- In the future, when we can support dynamic / additional columns, it'd be nice to have a column showing the version from each channel and a column with the selected channel. That'd make managing things a lot easier for those interesting in keeping things on the latest version.
- In the absence of a separate column for release channel, would it be worth the space to have a small "badge" after the addon title or before / after the installed version to easily see what everything is set to? Should we always have that attribute available in the default view once this feature is shipped vs having it buried in the expanded row?

src/gui/update.rs
```diff
@@ -509,6 +538,37 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
             ajour.header_state.previous_sort_direction = Some(sort_direction);
             ajour.header_state.previous_sort_key = Some(sort_key);
         }
+        Message::ReleaseChannelSelected(release_channel) => {
+            log::debug!("Message::ReleaseChannelSelected({:?})", release_channel);
+
+            if let Some(expanded_addon) = ajour.expanded_addon.clone() {
```
tarkah, line 544: Unnecessary clone, use `&ajour.expanded_addon` instead. We don't need an owned version of this.

tarkah: If a type is not `Copy`, it will try to move that value / "take ownership" and the borrow checker will obviously complain. Always default to using the borrowed value and instead have the borrow checker tell you when you need an owned version. That should help prevent adding an unneeded clone just to get the borrow checker to stop complaining.

tarkah: What editor do you use? Are you using Rust Analyzer? I find the `inlay hints` it provides extremely helpful in cases like this. For example here's what VS Code shows me before and after. Notice the `inlay hint` of `Addon` before and `&Addon` after. It makes it really easy to see that we were using an owned value and now a borrowed value.

image

image

casperstorm: Using Vim, but with Rust Analyzer. Will see if I can add the inline hits like that as well.

Screenshot 2020-09-26 at 07 43 01

casperstorm:
> * If you switch to Alpha, it shows an "Update" button. However, if you update and want to switch back to Stable, no button is available to switch. I  guess "Force Update" handles this, but it's not immediately apparent. Maybe for release channel changes, we present a button next to the channel picklist such as "Switch" when a new channel is selected that is available? This will also prevent the expanded row from collapsing if you click "Update" on the main addon row which feels weird.

Hmm good point, and great catch. I'll think a bit about this case and how we can solve this in a elegant way.

> * Should we show release channels in the pick list if they aren't available for the addon?

No we shouldn't – should be easy to filter out. 

>   * If we don't change this, then a UX issue I find annoying is that when you select an channel with no release, the "Release is X ago" text dissapears and the channel picklist / entire UI shifts up. Maybe we move that text to the right of the picklist?

Solved by the above, but still agree we might want to redo the UI a little.

> * I noticed a lot of the time, the Alpha or Beta channel are far behind Stable. I'd imagine most people are only interested in those channels if they are newer than the latest stable release? It'd be nice to know when the Alpha / Beta channel for an addon is "outdated" to stable if currently selected.

Yeah would require us to set some limits I guess.
I feel like its an advanced features where if we provide good enough feedback about HOW far they are behind we don't need to filter any out.

> * Likewise, it might be good to know when a new Alpha / Beta release is available compared to the installed Stable addon.

Hmm, good point. Need to rethink UI a little bit.

> * I guess for the previous two points, its actually hard to make these determinations since release date isn't helpful if the older Beta or Alpha release are still a "newer" more bleeding edge version being tested, so maybe we can't really do much here? It's not like everyone is using semantic versioning :P

I don't want to make this a feature everyone should use, becuase it IS advanced. But on the otherhand it needs to be useful, right.

> * I assume this also makes it difficult to have a sensible "default" release channel that isn't stable? Are you still planning on adding a "Default" release channel that can be specified in the settings, or a global "override all applicable addons to release channel X" type feature?

No, I dont want to have a "overwrite" button I think. I can't see the use-case to be honest. 

> * Would it be worth the space to have a small "badge" after the addon title or before / after the installed version to easily see what everything is set to? Should we always have that attribute available in the default view once this feature is shipped vs having it buried in the expanded row? Or maybe a new column between Title and Local Version so it can also be sorted?

Yes! This is also something I need to rethink regarding the UI.

casperstorm:
This feature is starting to look nice, @tarkah.
I have made some logic to the release channels so if you selected `beta` but `stable` is newer, we should that. Same goes for `alpha`. This way we always show what is relevant. I've also modified the GUI a little bit.

I am still a little unsure if I want to show the release channel a column. Maybe it would fit in when we do optional columns. What do you think?

tarkah:
Great, I'll test this out later today.

I agree we don't need a separate column until we can add dynamic columns as most people wont ever switch channels. I'll try to play around with a maybe adding a small badge to either the title or local version.

tarkah:
My thought for a "badge" would be `Text` that is pretty small inside a `Container` with a border and rounded corners. With the text and border color being all one color.

tarkah:
I'm not sure I love the logic for retaining the package as Stable over Beta or Stable / Beta over Alpha. It's really confusing when you switch channels on the addon, but the "is x ago" text doesn't change, nor does the remote version. It doesn't indicate why nothing is changing so it just seems like there is no channel available for the selection. Then it adds to the confusion when you have an addon that really doesn't have that channel available and it gives you the "has no available release" text and you wonder why the other addon didn't say the same thing.

I'd say we still "switch" the channel, but maybe we add a warning of "is x ago (older than Stable)" or something and leave it up to the user to switch it back. Maybe they want to keep it on that channel still.

tarkah:
> 
> 
> > * If you switch to Alpha, it shows an "Update" button. However, if you update and want to switch back to Stable, no button is available to switch. I  guess "Force Update" handles this, but it's not immediately apparent. Maybe for release channel changes, we present a button next to the channel picklist such as "Switch" when a new channel is selected that is available? This will also prevent the expanded row from collapsing if you click "Update" on the main addon row which feels weird.
> 
> Hmm good point, and great catch. I'll think a bit about this case and how we can solve this in a elegant way.

Have you thought of how you want to address this yet?

tarkah:
Wow, it's the first time I've messed with the frontend after the lib split and it compiles so much faster!! Makes messing with styles and layout so much more enjoyable :P

casperstorm:
I think it is a little bit confusing as well, but it is also the way the Twitch client does it. I think we just need to do some UI out of it. Also the way expected from this issue: https://github.com/casperstorm/ajour/issues/92
> Update to the latest Beta OR Release version of the addon. Whichever is newest.

If you select `beta` or `alpha` you expect to get the newest of the newest, no matter what channel otherwise it make no sense, this we achieve doing it this way.

casperstorm:
> > > * If you switch to Alpha, it shows an "Update" button. However, if you update and want to switch back to Stable, no button is available to switch. I  guess "Force Update" handles this, but it's not immediately apparent. Maybe for release channel changes, we present a button next to the channel picklist such as "Switch" when a new channel is selected that is available? This will also prevent the expanded row from collapsing if you click "Update" on the main addon row which feels weird.
> > 
> > 
> > Hmm good point, and great catch. I'll think a bit about this case and how we can solve this in a elegant way.
> 
> Have you thought of how you want to address this yet?

I think we need to remember it is a niche feature. Changing AND updating into another release channel is something very few will do. I think we should stick to `force download` but consider renaming it – or moving the button. What do you say?

tarkah:
Here's how I envision the badges working. I decided it works best to only show badges when "Alpha" or "Beta" is selected, otherwise it's noise with a bunch of "Stable" badges.

Do you like this?

image

casperstorm:
> Here's how I envision the badges working. I decided it works best to only show badges when "Alpha" or "Beta" is selected, otherwise it's noise with a bunch of "Stable" badges.
> 
> Do you like this?
> 
> image

Seriously, @tarkah – sometimes I feel like we are brothers from across the world. I had EXCATLY the same idea, was even now just looking to find such an example. YES I like that.

tarkah:
Woo! I'll open a PR so you can merge.

tarkah:
> 
> 
> I think it is a little bit confusing as well, but it is also the way the Twitch client does it. I think we just need to do some UI out of it. Also the way expected from this issue: #92
> 
> > Update to the latest Beta OR Release version of the addon. Whichever is newest.
> 
> If you select `beta` or `alpha` you expect to get the newest of the newest, no matter what channel otherwise it make no sense, this we achieve doing it this way.

Ahh, I missed the conversation that took place there. Yeah, the logic makes sense in that context.



> > > > * no button is available to switch. I  guess "Force Update" handles this, but it's not immediately apparent. Maybe for release channel changes, we present a button next to the channel picklist such as "Switch" when a new channel is selected that is available? This will also prevent the expanded row from collapsing if you click "Update" on the main addon row which feels weird.
> > > 
> > > 
> > > Hmm good point, and great catch. I'll think a bit about this case and how we can solve this in a elegant way.
> > 
> > 
> > Have you thought of how you want to address this yet?
> 
> I think we need to remember it is a niche feature. Changing AND updating into another release channel is something very few will do. I think we should stick to `force download` but consider renaming it – or moving the button. What do you say?

Honestly, it's probably fine the way it is. When I first used this, I was thinking the release channels were 100% isolated. So you'd always be on the beta / alpha release, not the "latest". Considering that, most the time someone reverts to "Stable", the local / remote will probably already be the stable version so there is nothing that needs to happen. So I think the "Force Update" usage is fine for that edge case where someone does want to move from a Alpha / Beta installed release back to Stable.

casperstorm:
I just slightly increased font size on badge @tarkah. Looks really nice tho, and I think it's a good idea to only show it when release channel is `beta` or `alpha`.

I'll go do a little more testing, but otherwise I think we should consider this feature done, what do you say?

tarkah:
Looks good to me! I think this is good to go.

Issue #128, opened 2020-09-24
Optimize rendering time, view, and layouting. 


casperstorm:
When running a debug build and hitting `f12` we are able to inspect different metrics from `iced`:

<img width="1307" alt="Screenshot 2020-09-24 at 15 41 23" src="https://user-images.githubusercontent.com/2248455/94153066-9d7f8000-fe7c-11ea-9816-e113598bcc99.png">

I think it would be nice to look into if we could get these numbers even lower. We don't have any performance issues, but we always want to be as performant as possible so I am sure there are optimizations to be done in out `update.rs` and `element.rs` file.

I don't see this as an important issue. I see it as a reminder to us.
Any comments on this, @tarkah?

tarkah:
Yeah this is something we definitely need to dig into at some point. I've done some rust profiling with `perf` before, but on much simpler programs. I wonder how well it'll work with the Iced framework making a lot of noise.

I have no context for how fast each of these metrics should be, plus we are running a debug build so it'll be much lower than the release profile. But it's a good benchmark to compare versions or new features to see if things slow down at all. Or when figuring out optimizations, if it improves any of them.

I'll definitely start utilizing this when adding features to make sure nothing impacts performance over the current release. And I'll try to play around with `perf` or other profiling tools to see if we can uncover any expensive operations that we can focus on.

tarkah:
Interestingly, on windows, the Update speed is orders of magnitude faster! I average somewhere around 500-600ns. Layout is a bit quicker too. I wonder what the main driver for that difference is. I assume those operations are completely CPU bound?

![image](https://user-images.githubusercontent.com/10239377/94168851-eb779200-fe42-11ea-9fa0-16d098026354.png)

tarkah:
Or could just be when you took the screenshot... It appears the Update starts high when first launched, then drops pretty quickly after moving the cursor around for a bit.

casperstorm:
To be honest I haven’t debugged it TOO much. It’s just a thing I had in my mind for some time that it would be nice to do some debugging on.

Also we have this in our main.rs:
https://github.com/casperstorm/ajour/blob/a515c4c84e87eee29e92a3bbb8a54516951074fa/src/gui/mod.rs#L347

tarkah:
> Also we have this in our main.rs:
> 
> https://github.com/casperstorm/ajour/blob/a515c4c84e87eee29e92a3bbb8a54516951074fa/src/gui/mod.rs#L347
> 
[...]

PR #130, opened 2020-09-24
split ajour into bin & lib

tarkah:
split out ajour into a lib so we can benchmark / profile functions easier without the gui


tarkah:
as discussed in #128, this is the start of moving ajour to a lib. Anything we want to reference we can throw in that `exports` module. Check out the bin `parse_addon_directory` I added. Now that I have this, I'm going to look at running the profiling for that specific function.

tarkah:
So I think its best if we actually take all the non-gui stuff and push it into its own crate / own dependencies. Right now all the `bins` I'm creating, you still have to compile Iced which is super annoying. If we just leave the `gui` in the main ajour crate, then split everything else out to an `ajour-core` crate, we should have better compilation times when not touching the frontend at all. What do you think?

tarkah:
Well... here's what that would look like :P I split out everything except for the `gui` module into a `ajour-core` crate. Since we store `button::State` in the Addon, I added a feature flag `gui` on the core crate, and made those fields / the `iced_native` dependency  conditional on that feature flag. Now we can use `ajour-core` in other bins without any dependency on iced / other gui libs.

I realize this is a HUGE structural change, so I completely understand if you do not want to make this move.

casperstorm:
I've been going through this, @tarkah and it is the way forward.
This is much more maintainable, faster compile time, and easier to test.

The plan will be the following:
I'll finish #129 and merge it to master. Then we will do some proper testing on master and make a release.
Meanwhile, this branch is finished, and master is back merged into it so we are ready to merge this and going forward using this structure. This is a huge improvement, and yes it is a big structural change but a needed one with the amount of features we keep doing. Good job mate and keep working on this.

casperstorm:
@tarkah I'll finish my branch today or tomorrow. :)

tarkah:
Great, glad you are aligned on this change! Sounds like a plan. Can't wait to see the release channels get finished up, that'll be a good addition. It's going to be a big release!

tarkah:
I've rebased to master, this should be good to merge!

tarkah:
I added a couple more `bins`... `parse_toc` and `fingerprint_addon`

PR #133, opened 2020-09-28
fix: missing dep

tarkah:
Oops, I did the lib split on my linux desktop and accidently removed `dirs` as a dependency due to conditional compilation. Just caught it this morning on my windows work computer.


casperstorm:
Good catch.

PR #167, opened 2020-09-30
feat: better cache busting

casperstorm:
## Proposed Changes
Ajour will now compare modified date to determine if we should rehash an addon.
Can you look this through @tarkah.

## Checklist

- [X] Tested on all platforms changed
- [X] `cargo fmt` has been run on this branch
- [X] `cargo clippy` has been run on this branch

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 
 ### Changed
 - Reworked the controls for selected flavor. It should now be more obvious what is selected and what you can select.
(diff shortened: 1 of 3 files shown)


tarkah:
So on windows at least, the modified date of the folder doesn't update when you edit the `.toc` for example, it only appears to change during the following:

```
The folder's last modified time is updated for these actions:

   - new file or folder directly in target folder
   - renamed file or folder directly in target folder
   - deleted file or folder directly in target folder
   - hardlink create/delete/rename - same as files
   - file/folder symlink create/delete/rename
   - directory junction create/delete/rename
```

So this isn't _perfect_, but it should definitely bust the cache when users delete / reinstall addon folders, which I think is mostly what we're trying to solve

PR #168, opened 2020-09-30
include ajour_core in logging

tarkah:
@casperstorm I think this was the regression you were referring to. When we created `ajour-core`, we never updated our log filter to capture log events from that library


casperstorm:
Good catch.

PR #181, opened 2020-10-01
feat: catalog

tarkah:
will resolve our number one feature request!

currently just has the catalog data model and API request added. @casperstorm lets both work off this branch


casperstorm:
Small thing i noticed:
* You can select "none" in categories after you have selected on. Don't know how we should handle this best?

casperstorm:
@tarkah, this is what I see we are currently missing for a initial release:

Ajour:
* Small design tweaks.
* Handle flavors properly in client (they are added to the struct: https://github.com/casperstorm/ajour/pull/181/commits/ff15da79344c54ab7193f5f3bfa10dc3dc82c1c0)
* Two buttons in list (Retail, Classic)
* Review https://github.com/casperstorm/ajour/pull/192
* Nice to have: round numbers to SI (1,000,000 > 1M)
* Nice to have: picklist with different sources
* Nice to have resizable columns

API:
* Nice to have: tukui api
* ~~Send flavors along in api~~

tarkah:
@casperstorm I think this is in a good state to review! It needs a little love on the UI, but I know you will do a good job there :)

Please review the logic from the user clicking the "Install" button all the way to it getting unpacked. Here is a high level overview:

- User clicks install button for either flavor
- We call the relevant API to get information about the addon
- I create an `Addon` using that information and insert it into the relevant flavor of `Vec<Addon>` in `Ajour`. For curse addons, I use the latest file matching the flavor that is stable.
- We set the `Addon` to "Downloading". This will cause the relative button in the catalog to become disabled and show "Downloading".
- At this point, we call `perform_download_addon` and the `Message::DownloadedAddon` message, which then all works identically as when an update is available, from downloading to unpacking.

casperstorm:
Screenshot 2020-10-05 at 08 54 20

I've tidied up the buttons, titles and so on. It's starting to look good.

casperstorm (approved):
@tarkah can you take it for a final spin?

tarkah:
@casperstorm Looks great to me! I cleaned up a few small things, mostly naming. I think we are ready to ship this!

Issue #162, opened 2020-09-30
Linux build


Ishayu:
**Is your feature request related to a problem? Please describe.**
I didn't know how to install it on Manjaro (which is based on Arch Linux).

**Describe the solution you'd like**

```sh
sudo pamac install rustup 
```

If it points out there's a conflict with rust, install rustup anyway, removing rust.

Next, install a toolchain.

```sh
rustup install stable
```

```sh
rustup default stable
```

Then follow the standard build steps.

**Additional context**
By default all the necessary libraries seem to be installed, at least on Manjaro. This isn't too surprising as you do a lot of self-compiling on Arch and derivatives. Not 100% certain though, so a review might be good.

Can't figure out how to send the branch for a pull request. I just get a 403. But you can have this, hope that's OK.

tarkah:
What error are you running into when you `cargo build --release` ?

Ishayu:
rustup is not installed. This is a feature request to add that instruction to the readme file, not a bug. :)

It's a brilliant program and it works well on Linux. Good job!

gageracer:
you can build it using aur

fishxz:
how about a flatpak build?
would be the best way to cover linux

 https://flatpak.org/

tarkah:
I've added an `AppImage` release #189 . I'll try to get the AppImage for 0.3.5 published to that release once @casperstorm is back online. I can link it in discord in the meantime

masonnl:
fwiw, on Manjaro I was able to build from source using rustup (v1.22.1-1 [Official Repositories]) and its optional dependencies (lldb and gdb. The final bins installed for me looked like this:

![step 3](https://user-images.githubusercontent.com/4482337/94983724-ccdc5f80-0502-11eb-8ad4-0fa588d5978d.png)

Excited to have an AppImage! Thanks @tarkah

tarkah:
Closing now that we have AppImage on our 0.3.5 release

https://github.com/casperstorm/ajour/releases/tag/0.3.5

PR #179, opened 2020-10-01
add snap package

tarkah:
partially satisfies #137 

## Proposed Changes
  - Provide a `.snap` that linux users can install

## Checklist

- [ ] Tested on all platforms changed
- [ ] `cargo fmt` has been run on this branch
- [ ] `cargo clippy` has been run on this branch
- [ ] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users

PR #189, opened 2020-10-03
add appimage release

### LINKED ISSUE #137: Provide Ubuntu/Linux port
I am playing WoW with lutris/wine which works perfect. 
I tried running your .exe with wine but it crashed

Could you provide a port of your app for Ubuntu/Linux?

Make it work with wine but I would clearly prefer a real port. Maybe you could provide a snap or deb installation file

tarkah:
Resolves #137 #162

## Proposed Changes
  - release AppImage that can be run on any linux distro

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 ### Packaging
 - Updated Ajour icon.
 - Added an alternative build that uses OpenGL. This will allow Ajour to be used by the widest possible audience, and resolve issues where users couldn't use Ajour with older / certain GPU configurations. An alternative download link will be provided to users wanting to try this build over the default.
+- Added AppImage release that can be used on linux distro
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


tarkah:
@casperstorm message me when you're online, I created an AppImage for `0.3.5` that we can publish to that release. I'm not too familiar with Makefile format, but I added a line in there to create the AppImage. I can provide this for future releases until we get it integrated into our github actions. I'm going to close out #179 for now, since I don't think it's necessary anymore.

I've tested this on a fresh Ubuntu 20.04 live image, and a ~1year old live image of Solus that I had laying around. Both worked out of the box perfectly!

casperstorm:
@tarkah I think this is looking good.
Only question is if we should rename `app` in make to something else so its releated to macos. `app` and `appimage` is a bit confusion now. Other than that this is approved.

tarkah:
`.AppImage` is the official extension.. but maybe we can make it like `Ajour-Linux.AppImage`?

casperstorm:
> `.AppImage` is the official extension.. but maybe we can make it like `Ajour-Linux.AppImage`?

Looking at it again i think its fine tbh.  
`app` does generate a `.app`
`appimage` does generate a `.appimage`
It was just - you need to know to know right? But I think its fine, it will be bound to CI soon so its out of human touch.

https://github.com/casperstorm/ajour/blob/a9ffcb0f582d9552a8886e65cad3ceb9b10d591e/Makefile#L62

https://github.com/casperstorm/ajour/blob/a9ffcb0f582d9552a8886e65cad3ceb9b10d591e/Makefile#L39

But looking at it again it's

PR #192, opened 2020-10-03
feat: abstracted all categories option for a cleaner interface

kasperkronborg:
## Proposed Changes
  - Abstracts away the very implicit api of injecting the All Categories into the category dropdown.
  - The new interface implements both `Default` and a static function for retrieving the `all_option()` and it resembles the interface of the result size dropdown to generate a more cohesive codebase.

### DIFF
--- a/crates/core/src/catalog.rs
+++ b/crates/core/src/catalog.rs
@@ -46,11 +46,37 @@ pub struct Catalog {
 pub struct CatalogAddon {
     pub id: u32,
     pub name: String,
-    pub categories: Vec<String>,
+    pub categories: Vec<CatalogCategory>,
     pub summary: String,
     pub number_of_downloads: u64,
     pub source: Source,
     pub flavors: Vec<Flavor>,
+}
+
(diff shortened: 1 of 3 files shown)


casperstorm (approved):
Normally I don't think its necessary to replace a string with a struct if it does not have a purpose for being a struct However I think it can be justified in this PR because the end product here is more readable than before.

kasperkronborg:
@casperstorm optimized it ever further, so I no longer transforms the category string directly on the catalog struct, but's now all handled in the ui part of the application.

casperstorm:
> @casperstorm optimized it ever further, so I no longer transforms the category string directly on the catalog struct, but's now all handled in the ui part of the application.

Nice, this is better. 👍🏻

tarkah:
👌 nice!! This is perfect. We need to get ResultSize working directly in the picklist as well instead of the String of it. That's so much cleaner!

tarkah (approved):
Looks great! Love the change to enum, this feels a lot better.

PR #204, opened 2020-10-06
Proposal : Add localization

psionikangel:
Hi there!

This is a proposal PR to add localization to the ajour app that I want to bring up for discussion. This is a big change and it most likely would affect any other future development because from the day you add localization, it always must be taken into account moving forward. So this is a tentative PR to see if that is something that you want to consider adding to Ajour.

I want to start right from the start by saying that I'm completely new to rust and so the code might not be very standard or efficient, but it worked well enough to warrant at least showing it so you can look and take a decision on whether or not you want it in your app.

I have used the `i18n-embed` and `tr` crates with the `cargo-i18n` and `xtr` tools to achieve the translation. All that is based on the well known **GNU gettext** toolkit for software translation. The added code works out the locale of the system and then presents the translated strings to the user based on that. (I.E. As opposed to a settings dropdown letting you choose between different languages). There are of course pros and cons to both approaches, but the one I chose was the simplest to implement since there is no intervention from the user.

It works for most of the labels of the application. Of course, it is kind of a pain to maintain because all the strings presented to the user are not in a central location, so the workflow process to keep everything up to date means scouring through the entire codebase looking for strings and encasing them inside the `tr!()` macro. For now, there's not a lot of text so it kind of work out all right. Of course, the most efficient approach, if you do want to consider adding translations to the app, would be to eventually centralize all user-facing strings so they are easily translated.

This is what this looks like on my Mac set to a french locale, with the strings translated.

Ajour translated

Ajour translated

Small caveat : On my Mac I can't figure out, for the life of me, how to make sure the translations get included in the Ajour.app content. The binary release from running `make binary` works and is translated, but not the `.app` file.

So let me know what you think of this. Looking forward to discuss this proposal with you.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@
 
 # Editor config
 .vscode/
+
+# Translation files exclusions
+/i18n/pot
+/i18n/mo
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 
 ## [Unreleased]
 
+## [0.3.5] - 2020-10-01
 ### Packaging
 - Updated Ajour icon.
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -61,7 +61,6 @@ dependencies = [
  "iced_native",
  "image",
  "isahc",
- "locale_config 0.2.3",
  "log",
  "log-panics",
  "native-dialog",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -55,16 +55,20 @@ dependencies = [
  "chrono",
  "embed-resource",
  "fern",
+ "i18n-embed",
  "iced",
  "iced_futures",
  "iced_native",
  "image",
  "isahc",
+ "locale_config 0.2.3",
  "log",
  "log-panics",
  "native-dialog",
  "opener",
+ "rust-embed",
  "timeago",
+ "tr",
  "widgets",
 ]
 
@@ -412,6 +416,15 @@ name = "cloudabi"
 version = "0.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ddfc5b9aa5d4507acaf872de71051dfd0e309860e88966e1051e462a077aac4f"
+dependencies = [
+ "bitflags",
+]
+
+[[package]]
+name = "cloudabi"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4344512281c643ae7638bbabc3af17a11307803ec8f0fcad9fae512a8bf36467"
 dependencies = [
  "bitflags",
 ]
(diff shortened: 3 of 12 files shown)


casperstorm:
What an amazing first contribution @psionikange 🎉
We have discussed it a few times and was preparing to do a localization feature so great timing!

I really like this approach it is very smooth adding a new language, just tried on your branch. 
Also kudos for adding a the descriptive `TRANSLATE.md` file.

Here are my initially thoughts:

1. I would like a manual overwrite function in settings.
    1. Ideally I would like a picklist underneath `Themes` in Settings where the user can select the language they want.
    2. This value should properly be saved to `Config`.
2. Avoid trying to detect language based of locale but rather always (until user picks something else) show `en`.
3. Update `README.md` with a field like where you link them to the `TRANSLATE.md`. Would fit nicely under [Contribution](https://github.com/casperstorm/ajour#themes).
4. Update `CHANGELOG.md` with relevant imformation about localization possibilities, this relates to 1. if we do a picklist.
5. Run and correct `clippy` issues.

How does this sound?

psionikangel:
Fantastic that you liked it. In this case, I will continue working on it to take into account your points and keep you updated on my progress through this PR's discussion.

casperstorm:
> Fantastic that you liked it. In this case, I will continue working on it to take into account your points and keep you updated on my progress through this PR's discussion.

Sound awesome.
And please ping me here, or on [Discord](https://discord.gg/4838t9R), if you have any questions.

casperstorm:
@psionikangel are you still working on this?

psionikangel:
Yes, I'm still working on it. I managed to add the pick list in the settings for the languages. Now I just need to figure out how to apply the translations on the fly when you change the settings.

casperstorm:
> Yes, I'm still working on it. I managed to add the pick list in the settings for the languages. Now I just need to figure out how to apply the translations on the fly when you change the settings.

Nice! Looking forward to see where this ends.

psionikangel:
Hi @casperstorm 

I've got some time to work on this and I'm in a bit of a jam trying to implement dynamic translations. I hope I can pick your brain and get your thoughts on my issue.

Adding the pick list for languages in the settings was easy enough by copying the behavior for the theme pick list. However, now I'm at the part where I need to redraw the whole UI so the text of every elements is translated when the user select another language. That's where I don't know exactly how to achieve that.

The library (i18n-embed) uses the macro tr!("some text") to translate the text of an element based on the requested languages. What I originally did was to place that macro everywhere a UI element was created like so :

```
// Directory button for Backup directory selection.
        let directory_button_title_container =
            Container::new(Text::new(tr!("Select Directory")).size(DEFAULT_FONT_SIZE))
                .width(Length::FillPortion(1))
                .center_x()
                .align_x(Align::Center);
```

However, that means the text <Select Directory> will be translated only when the element is created, as the text is not part of the state.

So I guess here's the question :

Do we solve this by making the display text of every UI element part of the state so it can be dynamically changed when the language is changed in the settings? That seems like a very large codebase change that I don't know if I'm comfortable to make.

I also checked if the UI library, **iced**, was providing some higher level feature to enable translation and it doesn't seem to be the case, which means that If we want something dynamic, I guess the only way is to make the display text part of the state and the update function for the language pick list iterate over every UI element to change their states.

Let me know what you think.

casperstorm:
I am a little confused because it should do a re-render everytime the `fn view` i called.
But that does not work or?

casperstorm:
I've marked it as draft. You can change that once it's ready @psionikangel.

casperstorm:
@psionikangel how is the work going here? 🙂

psionikangel:
Hi @casperstorm 

I didn't have much time to work on it. I did try something just now based on my [discussion thread](https://github.com/kellpossible/cargo-i18n/issues/41) with the cargo-i18n library author, but It doesn't seem to work. I may very well be out of my depth here as I can't figure out how to translate everything on the fly.

While experimenting with the lowest amount of code needed to translate the app, I have come to the conclusion that unless the translation happens before the gui is started in `main.rs`, nothing gets translated.

```
// Start the GUI
    gui::run(opts);
```

I have experimented by placing the translating code at the top of `pub fn settings_container<'a, 'b>(` in `element.rs` (that doesn't work either) and at the top of `fn view(&mut self) -> Element<Message> {` in `mod.rs` thinking that the view function would re-render the app with the translations but no luck.

tytannial:
A font changer needed. It seems the default font can't show cjk font.

casperstorm:
> A font changer needed. It seems the default font can't show cjk font.
> 

Was this meant as a seperate issue?

tytannial:
> > A font changer needed. It seems the default font can't show cjk font.
> > 
> 
> Was this meant as a seperate issue?

If the font is part of the localization or the theme. idk. 🤣

casperstorm:
@psionikangel I am wondering if we are blocked by this: https://github.com/hecrj/iced/issues/103#issuecomment-561464214

psionikangel:
Hi @casperstorm,

I have to say that I did search about localization issues in the iced repo at first but I did not find this particular issue, so congrats for finding this. To answer the question, I do think that yes, it definitely will be blocking. (Maybe not for french, but for several other languages where we will definitely depend on UTF-8/different fonts being available to translate properly).

So all things considered, maybe we shouldn't implement a solution that will not allow us to at least translate in all the languages World of Warcraft is available in.

Also, ideally, as is suggested in https://github.com/hecrj/iced/issues/103#issuecomment-561464214 if the localization support was directly supported by the UI framework, that would immensely simplify the work that we have to do.

It is kind of sad that as the iced author mentions, it seems that the overall support for internationalization fonts doesn't seem very mature in rust yet, so I guess all there is to do is wait.

casperstorm:
> I guess all there is to do is wait.
I agree.

casperstorm:
I has now been implemented https://github.com/casperstorm/ajour/pull/473.

PR #212, opened 2020-10-07
Ci/release workflow

mlablah:
Resolves #

## Proposed Changes
  - This adds a workflow that uses the workflow_dispatch event, meaning it can be triggered manually via the Actions tab.
  - It will use the makefile as it works at the moment on the ci/makefile branch to build all 6 assets (mac, win, lin x vulkan, opengl)
  - If the builds are successful it will create a new draft release and tag. And upload the 6 assets to it. 
  - It is still necessary to update the version in cargo and changelog.md prior to running the workflow, and to add the changelog section to the release description before publishing it.

## Checklist

- [ ] Tested on all platforms changed. The builds have not been tested (only that the windows versions start). I noticed the macos builds generated are identical in size with and without opengl.

### DIFF
--- a/.github/workflows/release_to_draft.yml
+++ b/.github/workflows/release_to_draft.yml
@@ -1,4 +1,4 @@
-name: "Release matrix"
+name: "Release to draft"
 
 on:
   workflow_dispatch:
@@ -47,16 +47,10 @@ jobs:
 
       - name: "Windows wgpu?"
         if: ${{ matrix.os == 'windows' && matrix.gfx == 'wgpu' }}
-        uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: --release
+        run: make binary
       - name: "Windows OpenGL?"
         if: ${{ matrix.os == 'windows' && matrix.gfx == 'opengl' }} 
-        uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: --release --no-default-features --features opengl
+        run: make binary OPENGL=1
 
       - name: Upload AppImage wgpu
         if: ${{ matrix.os == 'ubuntu' && matrix.gfx == 'wgpu' }} 
@@ -87,13 +81,13 @@ jobs:
     needs: release
     name: "Create Release"
     runs-on: ubuntu-latest
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
     steps:
       - name: Create Release
         id: create-release
         uses: actions/create-release@v1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           tag_name: ${{ github.event.inputs.tag }}
           release_name: ${{ github.event.inputs.tag }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release_to_draft.yml
```diff
@@ -0,0 +1,164 @@
...
+        required: true
+
+jobs:
+  
+  release:
+    name: "Release"
+    strategy:
+      matrix:
+        os: [windows, ubuntu, macos] 
+        gfx: [vulkan, opengl]
```
tarkah, line 17: change `vulkan` to `wgpu` since vulkan is technically only used on linux, and metal for macos / directx12 for windows

tarkah:
@mlablah if you want to reduce the redundant tasks for the matrix strategy, an alternative way to structure this would be to have a "single" matrix, where each option has it's own variables that you can reference, such as:

```yml
   strategy:
      matrix:
        target:
          - name: Windows 
            os: windows-latest
            make_command: make binary
            gfx: wgpu
            binary_path: target/release/ajour.exe
          - name: Macos OpenGL 
            os: macos-latest
            make_command: make binary dmg OPENGL=1
            gfx: opengl
            binary_path: target/release/osx/ajour.dmg
          - name: Linux
            os: ubuntu-latest
            make_command: make binary appimage
            gfx: wgpu
            binary_path: ajour.AppImage
```

And then you can really have just 1 task to run the make command and one to upload the asset. You can then do the same thing on the side for uploading all the assets to the release. The "Create Release" job can be separate from the "Upload Assets" job, and actually depend on it and get the upload URL from it, so that the "Upload Assets" job can become a matrix strategy. You can see how I pass the upload URL between jobs in my actions file here:

- Define an output on the job: https://github.com/tarkah/tickrs/blob/d1bf56c7f31ea7ad96a5ee024ab58ebdc2f94ef6/.github/workflows/release.yml#L13
- Have the other job depend on the previous job: https://github.com/tarkah/tickrs/blob/d1bf56c7f31ea7ad96a5ee024ab58ebdc2f94ef6/.github/workflows/release.yml#L28
- Access the variable: https://github.com/tarkah/tickrs/blob/d1bf56c7f31ea7ad96a5ee024ab58ebdc2f94ef6/.github/workflows/release.yml#L101

mlablah:
> @mlablah Awesome stuff! If you want to reduce the redundant tasks for the matrix strategy, an alternative way to structure this would be to have a "single" matrix, where each option has it's own variables that you can reference, such as:
> 
I wanted to do something like that, so I could reference the different target variables (and not just the os name) within the steps, but the problem is we need to generate 2x3 assets, and your suggested matrix would only spawn 4 jobs. I also tried combining an "explicit" matrix for the 3 targets and gfx "list" with options [vulkan,opengl] but that isn't supported either. 

And the windows builds don't use make, the linux builds have extra steps (downloading linuxdeploy), the appimage recipe has two different output filenames etc. - That's why I ended up with all those conditional steps in the end ^^;

Regarding passing the output url: I wanted to build first, that way ensuring that all assets could be generated before creating the draft release. That's why create-release is dependent on the matrix job release and not the other way around.

.github/workflows/release_to_draft.yml
mlablah: Will rename!

tarkah:
I only used three in my example, but you can have as many entries in the matrix array as you want. So you could totally have all 6 with all the needed variables. And yeah, I guess windows would need its own since we don't use make there. Anyways, this is awesome work! Can't wait to have this integrated into the release process!

PR #214, opened 2020-10-08
Alter wording in readme

layday:
'Curated' would imply that the list is selective but it aims
to be exhaustive.


casperstorm:
Thanks @layday, good catch.

PR #219, opened 2020-10-09
feat: refactoring color themes

casperstorm:
## Proposed Changes
- Refactoring the `ColorPalette` struct so we can have more colors and make the overall of Ajour better.
- Add themes directly to Ajour.

## Checklist


casperstorm:
@tarkah mind taking this for a spin?

tarkah:
@casperstorm awesome stuff! Overall a huge improvement, especially with anything lighter. I really love how Nord, Sort and Dracula look now. 

A couple preferences on my end:

- I think there should be more contrast between background and foreground colors. It seems our foreground color is darker in the default Dark theme now then before? See comparison below.
- I think in general text needs to have more contrast for many of the themes, which I believe is mostly driven by the `normal.surface` color?
- Alliance color is far inferior to Horde, but I guess there isn't much you can do there :P Kidding aside, the background should probably not be blue, or at least much darker for better contrast.

---

New dark is definitely "darker" and text is harder to make out:

- Previous dark:
ajour_u6ZG7LnlWO


- New dark:
ajour_jmNAnc0NyN

casperstorm:
@tarkah Good points, ill iterate a bit more on the themes. Good feedback.

PR #220, opened 2020-10-09
Fix/catalog no stable file

### LINKED ISSUE #215: Catalog Downloading state getting stuck
**Describe the bug**
Catalog - installing Addon with missing stable channel causing it to halt on "downloading" state

15:48:23.294 [ajour::gui::update][DEBUG] Message::ParsedAddons(classic, 52 addons)
15:48:28.732 [ajour::gui::update][DEBUG] Interaction::ModeSelected(Catalog)
15:49:37.538 [ajour::gui::update][DEBUG] Interaction::CatalogInstall(Curse, classic, 399223)
15:49:37.657 [ajour::gui::update][ERROR] No stable file found for curse id 399223

The addon i tried to install.
https://www.curseforge.com/wow/addons/softres/files

**To Reproduce**
Catalog -> Search "Softres" > Install the one with about 2k Downloads by clicking "Classic"

**Expected behavior**
Should leave downloading state

**Screenshots**
image

**Software involved**
Windows 10 - 2004
Ajour 0.4.0 Release

**Log Output**
[ajour.log](https://github.com/casperstorm/ajour/files/5348562/ajour.log)

tarkah:
Resolves #215 

It'd be nice to actually present an error to the user when an install fails. Maybe make the download button go red / disable with "Failed" and somewhere display a reason why ("No stable file available for addon"). But not sure the best way to handle that at this point. I'm happy leaving this PR open until we land on a good design choice for that

## Proposed Changes
  - If addon download fails, remove the "empty" placeholder addon so that it doesn't appear in "My Addons" and the install button reverts from being stuck on "Downloading"

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 
 ### Fixed
 - Fixed bug where orphaned folders could exist after updating an addon if the newer version of an addon didnt't include those folders anymore. 
+- Fixed catalog install buttons getting stuck when install fails or addon is unavailable to download. Button will now show "Retry" if failed and disabled as "Unavailable" if the addon is unavailable.
 
 ### Packaging
 - Added Forest Night theme
--- a/crates/core/src/curse_api.rs
+++ b/crates/core/src/curse_api.rs
@@ -163,7 +163,7 @@ pub async fn fetch_game_info() -> Result<GameInfo> {
     }
 }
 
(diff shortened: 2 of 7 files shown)


casperstorm:
How can we move this forward @tarkah? Do you want me to do something?

tarkah:
@casperstorm I've change the way catalog install flow works. Instead of inserting an empty addon, I manage a list of statuses that we can leverage for changing the catalog install button. This will now give the user a Retry button if the install fails, or a disabled Unavailable button if we can't find the addon from the API

Issue #53, opened 2020-08-27
Invalid Zip archive: Could not find central directory end


WeaponMan:
I have Questie v6.0.4 installed and I wanted to update it to the last version v6.0.5, but its ajour is downloading some botched zip that cannot be open, when I do it from the curseforge website manualy opening the zip works.

Error is "Invalid Zip archive: Could not find central directory end".

Every other addon I had updated succesfully, so maybe something wrong with the curseforge api?

casperstorm:
Thanks for reporting this, @WeaponMan.

I just tested and it worked here on macOS. However, I'll do some more investigation because I did see this error before. The curseforge api is not that stable, but if the https module says the download is succesful then and only then do i send it to be unzipped.

So a few questions, if you have the time:
* Can you upload the botched zip so i can investigate it?
* Can you reproduce it everytime?
* Have you seen this happen on other addons?

<img width="804" alt="Screenshot 2020-08-27 at 12 50 33" src="https://user-images.githubusercontent.com/2248455/91433732-07980b80-e864-11ea-9f17-762b3e12aafc.png">

WeaponMan:
- No need, zip from curseforge have 9.1MB, the downloaded "zip" by ajour doesnt get downloaded all the way randomly ending the stream early
- happens everytime
- I think this problem is only with big addons like questie, weak auras 2 (~11MB) does the same thing if I downgrade it manualy

casperstorm:
Hm, I think this could be the problem: https://github.com/casperstorm/ajour/blob/master/src/network.rs#L18.
I timeout the request if it last longer than 5 seconds, which... seems like a bug. I must have done a typo. Do you mind testing with a larger timeout to see if it fixed the issue?

WeaponMan:
It works now, I thought the timeout is not that "hard" so I didnt try to change that, atleast you could replicate with bigger addons.

Sh0ku:
I've got the same error with updating RaiderIO (Invalid Zip archive: Could not find central directory end).

My Logfile gives this messages:

> 10:44:02.978 [ajour::network][DEBUG] downloading remote version v202010010600 for RaiderIO
> 10:44:18.238 [ajour::gui::update][DEBUG] Message::DownloadedAddon((RaiderIO, error: false))
> 10:44:18.405 [ajour::gui::update][DEBUG] Message::UnpackedAddon((RaiderIO, error: true))

[...]

PR #221, opened 2020-10-10
Fix/check content length

tarkah:
Partially addresses #53, we still need a nice way of showing install errors in the catalog view.

## Proposed Changes
  - Check content length header of download and make sure response bytes length equals that, otherwise error out

## Checklist


casperstorm:
How can we move this forward @tarkah? Do you want me to do something?

tarkah:
@casperstorm I've rebased and pointed this at the branch for #220 since it needs to leverage those changes so the user will see a "Retry" button in the catalog when the download isn't successful (on content length). You can merge this after merging that one.

PR #222, opened 2020-10-10
Feat/self update

tarkah:
Resolves https://github.com/casperstorm/ajour/projects/2#card-46568938

## Proposed Changes
  - When update is available, user can click "Update" link and have the newest version downloaded. It will replace the current binary and relaunch from that latest binary

## Checklist

### DIFF
--- a/crates/core/src/utility.rs
+++ b/crates/core/src/utility.rs
@@ -71,7 +71,14 @@ pub async fn download_update_to_temp_file(
     bin_name: String,
     release: Release,
 ) -> Result<(String, PathBuf)> {
+    #[cfg(not(target_os = "linux"))]
     let current_bin_path = std::env::current_exe()?;
+
+    #[cfg(target_os = "linux")]
+    let current_bin_path = PathBuf::from(std::env::var("APPIMAGE").map_err(|e| {
+        ClientError::Custom(format!("error getting APPIMAGE env variable: {:?}", e))
+    })?);
+
     let current_bin_name = current_bin_path
         .file_name()
         .unwrap()
--- a/crates/core/src/utility.rs
+++ b/crates/core/src/utility.rs
@@ -1,7 +1,11 @@
-use crate::{network::request_async, Result};
+use crate::error::ClientError;
+use crate::network::{download_file, request_async};
+use crate::Result;
+
 use isahc::prelude::*;
 use regex::Regex;
 use serde::Deserialize;
+
 use std::ffi::OsStr;
 use std::path::PathBuf;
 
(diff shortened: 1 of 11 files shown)


tarkah:
Need to clean up a few things, but this works great on windows! I'll change the version back from `0.1.0` before merging, that's just to test the functionality.

We need to test on each platform, with each binary (regular vs `opengl`) to make sure it downloads the right file and works. This process currently will break if the user renames their binary, so we should add an Error message when that happens.

casperstorm:
I'll test on Mac. This will add SILO much value.

tarkah:
@casperstorm I believe this is ready. Please test again on mac. I'll change the version back before merging.

casperstorm:
Should we move this to draft again until we have solved the current problems, @tarkah?

tarkah:
Yeah, I'll change back to draft.

On windows, I'll probably have it launch the temp file, then add some logic to check it it's launched from the temp file. If so, it'll copy the temp file over the main binary, then relaunch from that. Then from the main, remove the temp file. I don't think this fix is needed for macos or linux.

Then for macos, we need to start publishing the `.app` which we can use for the self update process.

mlablah:
@tarkah Would this approach require any changes to how ajour is initialized on different OS'es? Sort of outsourcing the dirty work to a static binary, which then gets removed by ajour when its work is done.

```
ajour displays Update button which user clicks
ajour downloads ajour-updater to tmp location
ajour launches ajour-updater with parameters like location and name of current ajour binary 
ajour shuts self down
ajour-updater downloads latest version of ajour
ajour-updater replaces old ajour binary with new ajour binary (once the file isn't locked by running process)
ajour-updater launches new ajour with "just_updated" parameter 
ajour-updater shuts self down
ajour if launched with "just_updated" parameter, will attempt to delete ajour-updater from tmp location
```

casperstorm:
@mlablah thats actually a nice idea. @tarkah what do you think? Should this be the first repo on our new org? ;)

tarkah:
Yeah I originally used that self update crate but the internals were simple enough that I just removed it  and did the logic myself do we didn't have dependency bloat.

I think we can use the logic @mlablah describes with one binary still. If we add an undocumented flags for when we launch from the temp downloaded binary we can easily have it copy itself over the original one (which now isn't running) then launch it. Then we pass the just updated flag to that and it'll remove the temp binary. We can code all this into our Ajour binary without the need for a  launcher

tarkah:
@casperstorm all done! Please do one final review and you're good to merge.

crates/core/src/utility.rs
```diff
@@ -86,6 +214,36 @@ pub fn wow_path_resolution(path: Option<PathBuf>) -> Option<PathBuf> {
...
+
+    retry(Fibonacci::from_millis(1).take(21), || {
+        match fs::rename(from, to) {
+            Ok(_) => OperationResult::Ok(()),
+            Err(e) => match e.kind() {
+                io::ErrorKind::PermissionDenied => OperationResult::Retry(e),
+                _ => OperationResult::Err(e),
+            },
+        }
+    })
```
casperstorm, line 240: Nice!

casperstorm (approved):

src/gui/update.rs
```diff
@@ -1397,9 +1402,54 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
...
+                    let mut args = std::env::args();
+                    args.next();
+
+                    match std::process::Command::new(&temp_bin_path)
+                        .args(args)
+                        .arg("--self-update-temp")
+                        .arg(&current_bin_name)
+                        .spawn()
+                    {
+                        Ok(_) => std::process::exit(0),
```
casperstorm, line 1435: I love this line for :D

PR #226, opened 2020-10-11
feat: tukui addons

petergp:
## Proposed Changes
  - Uses `catalog` feed to show both curse and tukui addons in catalog
  - Append `X-Tukui-ProjectID` to `toc` file for Tukui addons with that field missing (only support for archives containing just 1 addon. Some archives from tukui are a collection of addons)

Most tukui addons work, but some doenst because the zip archive contains multiple addons and the main one doenst have the `X-Tukui-ProjectID` key set. I haven't found a reliable way where I can detect which of the addons in an archive with multiple that I should add the id to. My suggestion is that we blocklist addons that we know is not working in the catalog.

## Checklist


casperstorm:
We should get this back-merged so we can merge it together with #315.

tarkah:
@casperstorm I've enabled the Tukui catalog already in #315. I assume we don't want to modify `.toc` entries from this PR as we now have the cache?

casperstorm:
> @casperstorm I've enabled the Tukui catalog already in #315. I assume we don't want to modify `.toc` entries from this PR as we now have the cache? 

True. Forgot this was more than just enabling catalog.

casperstorm:
Closing for now in favor for the new caching system.

Issue #107, opened 2020-09-13
Elvui Add-on folder is deleted if an error is encountered during the update process.


hamdeli:
**Describe the bug**
 Specific add-on's app folder is deleted after failure to update due to permissions replacing existing files.

**To Reproduce**
Steps to reproduce the behavior\:
_Requires an older version of ElvUI to be manually installed so as to trigger the update availability prompt in Ajour._
1. Launch Ajour normally, no elevated permissions.
2. Take note of add-count.
3. Click to update ElvUI.
4. If error is encountered, "access denied 5" for example, add-on count decrements.
5. Add-on is no longer present in the addons folder of WoW.

**Expected behavior**
Add-on folder that failed to update should not be deleted. It should be rolled back to its current version.
According to the source, it may be working as designed in the current version.

**Desktop:**
 - OS: [Windows 10 Pro]
 - OS Version: [1909]
 - OS Build: [18363.1082] 

**Additional context**
ElvUI updates prior to using Ajour were managed manually (not part of curse-forge). This is only happening in retail version. I have only reproduced this issue with ElvUI in retail.

hamdeli:
Label should be changed to "enhancement"

mlablah:
Discussion:
This sounds like a request that failing updates are rolled back to their previous state. 
But since we don't touch the local folders until after we have successfully downloaded the update, and extracted the content of the archive to a temporary location, this is about handling errors after this point.
Perhaps we could rename the current folders instead of replacing them, and keep them around until after the update finishes?

Also:
I just did a quick test to see if we have issues updating ElvUI specifically. 

To do that I downloaded the previous version of ElvUI by editing the current download URL on the tukui website
From: https://www.tukui.org/downloads/elvui-11.52.zip
To: https://www.tukui.org/downloads/elvui-11.51.zip
I then used Ajour to update it to version 11.52 which worked out.

@hamdeli When the update failed, did you notice if anything was left of the old version? Like a few files that it was unable to delete. Or was the ElvUI and ElvUI_OptionsUI folders both gone?

hamdeli:
@mlablah ElvUI folder was gone, but ElvUI_OptionsUI was still there. I will also add that I am running app version 0.3.3.

casperstorm:
> @mlablah ElvUI folder was gone, but ElvUI_OptionsUI was still there. I will also add that I am running app version 0.3.3.

[...]

PR #235, opened 2020-10-13
refactor addon and addon parsing

### LINKED ISSUE #233: Fallback strategy to improve addon recognition
Currently Ajour does not handle local addons that do not return a match from either CurseForge fingerprint endpoint or failing that, using the X-Tukui-ProjectID. This leaves us unable to show non-tukui addons that have been modified locally, or that have received updates not yet posted on CurseForge. 

 - If fingerprint endpoint doesn't return match: Use Curse ID from toc. Leave local version blank and give user `Repair` button to reinstall addon with latest from source. 
 - For remaining unrecognized folders, list them 1:1 and advice user to find and reinstall them through Catalog.
 - When installing addons through Catalog, make note of folder names contained in archive and correlate that with id from source. Use that correlation going forward if fingerprinting fails to return a match.

tarkah:
Resolves #233 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 ### Added
 - When pressing on either `local` or `remote` version in MyAddons you will see the changelog
 - When pressing on the addon title inside the catalog Ajour will open the addon website
+- Add fallback measures for displaying addons when fingerprinting fails or we can't link an addon to a remote registry
+  - Curse addons that have been locally modified should now display properly in Ajour. A `Repair` button will be present which will install the latest version of the addon so Ajour can accurately track the addon without local modifications.
+  - Addons that can't match to any registry will now show up in Ajour as status `Unknown`. Addons that have multiple folders will not be grouped and instead we will show one entry for every folder.
 
 ## [0.4.1] - 2020-10-11
 ### Added
--- a/crates/core/src/addon.rs
+++ b/crates/core/src/addon.rs
@@ -401,14 +401,14 @@ impl Addon {
     }
 
     /// Returns the version of the addon
-    pub fn version(&self) -> Option<String> {
+    pub fn version(&self) -> Option<&str> {
         if self.repository_metadata.version.is_some() {
-            self.repository_metadata.version.clone()
+            self.repository_metadata.version.as_deref()
         } else {
             self.folders
                 .iter()
                 .find(|f| f.id == self.primary_folder_id)
-                .map(|f| f.version.clone())
+                .map(|f| f.version.as_deref())
                 .flatten()
         }
     }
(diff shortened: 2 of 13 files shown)


tarkah:
Ok... this is going to be a pretty major PR. I decided to tackle refactoring `addon.rs` and `parse.rs`. I haven't touched `ajour` / gui yet, but I am able to test these changes currently with `cargo run --bin parse_addon_directory -- <path>`. I need to install some tukui addons to fully test.

I think it's really important to make the distinction between the following:

- `AddonFolder`... the actual folders and the `.toc` metadata we are able to parse from them. Nothing else. So basically just what lives on the filesystem.
- `RepositoryMetadata` - The information we can get about an addon from the repository
- `Addon` - this is kind of an abstract thing that doesn't actually _live_ on somoeones system. It is the glue that combines the `RepositoryMetadata` and bundles `AddonFolders` together to make an `"Addon"`. And where we store any other state / user defined preferences for the addon.

From this, I feel I am able to reason about everything a lot better. We parse all of the folders into `Vec<AddonFolder>`. Then we fingerprint all folders and extract any relevant repo identifies from them. From here, we can start querying the different APIs and generating our actual `Addon` structs on the relevant info.

As far as getting information from the `Addon`, I am planning on setting up `getter` methods that will work as follows. If we have `RepositoryMetadata` for an addon, it will try to get the value from there. If not, it will try to get that same metadata from the primary `AddonFolder` designated for the addon.  The `primary_folder_id` is intended to work just like the old `id`, as in it can be a unique identifier, but it also determines which `AddonFolder` we will try to query as a backup to `RepositoryMetadata` for values. When building an `Addon` for Curse with `RepositoryMetadata` and the slice of `AddonFolder`that matches the curse file module names, I just take the first folder name alphabetically (that appears to match how it's currently done). For Tukui, we just use the `AddonFolder` which had the tukui id.

Hopefully this all makes sense as you read through the diff and you think it's the right approach!

Note that I haven't actually tackled the fallback strategy yet, I'll add that next. Just wanted to get this major refactor out of the way first.

casperstorm:
While working on #107 I noticed something which we might wanted to handle now you are refactoring.
Consider the following:

```
15:37:57.419 [ajour_core::parse][DEBUG] retail - 1 addons with tukui id
15:37:57.532 [ajour_core::parse][DEBUG] retail - 0 addons applied with tukui id package metadata
15:37:57.532 [ajour_core::parse][DEBUG] retail - 1 unique fingerprints to check against curse api
15:37:57.705 [ajour_core::parse][DEBUG] retail - 0 exact fingerprint matches against curse api
15:37:57.705 [ajour_core::parse][DEBUG] retail - 0 addons applied with fingerprint metadata
```

In this case we see we have 1 addon with a tukui, but we didn't find any in the tukui api, and we didn't find any in the curse api.
So how could we debug this better? Right now we can't actually see that the problem is because the tukui api is down.
Just a thought for improvement.

tarkah:
Yeah that relates perfectly to the API issues going on. Without any data from the API, we either don't show the addon and show an error message, show the addon as "unknown" and 1:1 to AddonFolder, OR implement some cache we can fallback on and an offline status.

tarkah:
@casperstorm I feel this is in a pretty good spot. 

A big gap I caught that was introduced by this PR that I just fixed is around installing an Addon from the catalog. If a user installs an addon from the catalog, it was getting stuck on fingerprinting since the `AddonFolder`s of the addon weren't set. I've added logic to parse the `AddonFolder`s during the unzipping process and return them. Then if that download / unpack process was triggered due to the Catalog install, I'll use those addon folders to properly set the `folders` and `primary_folder_id` of the addon after unpacking finishes and before fingerprinting starts. This also fixes a gap if a user wants to delete the addon through Ajour after installing it from the catalog, it will now delete the folders since they're set on the addon.

It's worth noting that because of the spotty-ness of the fingerprinting API currently, certain addons (Details! specifically), are not getting returned. So when you `Refresh` a bunch of times, that addon goes from good to `Repair` and back and forth. That will obviously be confusing for people, but I guess less confusing than before where it just didn't show up :P Until we implement a local cache to fallback on, this will be a problem when the APIs suck.

I also rebased to Development and got the awesome changelog stuff working here :)

casperstorm:
Impressive, @tarkah. This is a giant leap forward.
I've tested, and for once it actually is kinda funny that the API is wonky so I easily cant test different states.
It's awesome. I tested it quite a lot and didn't find any problems. Went through code and looks clean as well. Much better than what we came from.

I just updated two with two commits:
- Removed the old dependency linking: https://github.com/casperstorm/ajour/pull/235/commits/b8f9f91520b673ce8784a5b90e02c926887a472c
- Renamed the `download_directory_for_flavor`: https://github.com/casperstorm/ajour/pull/235/commits/a7dde8fb5ad54acc1a7a6699e08209d57cbc58a4

I'll merge this to development and then we can do some testing on that branch.

Good job, man. Really good job.

PR #253, opened 2020-10-15
feat/changelog

### LINKED ISSUE #93: Read addon changelogs before updating
I really prefer to read the changelog of every addon before I update it so I know what is actually changing.

I can think of a few different ways this could be done. I'll list them in my preferred order, but happy for your expertise
1. If an update is available, when you click on the Remote version number the changelog will display in the expanded window
2. When you click on the addon name or remote version number the changelog will also show up underneath the summary
3. When you click on the addon name a Changelog button next to the other buttons (Force update/Ignore/Delete) that when clicked will bring a popup or expand further to show the changelog.
4. When you click on the addon name a Changelog button next to the other buttons (Force update/Ignore/Delete) that when clicked will load the changelog/addon website in your selected browser

N/A

casperstorm:
Resolves #93

## Proposed Changes
It is now possible to press on `local` or `remote` version to view a changelog.

Screenshot 2020-10-15 at 19 58 10 

## Checklist

### DIFF
--- a/src/gui/mod.rs
+++ b/src/gui/mod.rs
@@ -5,7 +5,7 @@ mod update;
 use crate::cli::Opts;
 use crate::VERSION;
 use ajour_core::{
-    addon::{Addon, AddonState, ReleaseChannel},
+    addon::{Addon, AddonState, AddonVersionKey, ReleaseChannel},
     catalog::get_catalog,
     catalog::{self, Catalog, CatalogAddon},
     config::{load_config, ColumnConfigV2, Config, Flavor},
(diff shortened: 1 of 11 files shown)


src/gui/mod.rs
```diff
@@ -62,9 +62,10 @@ impl std::fmt::Display for AjourMode {
 }
 
 #[derive(Debug, Clone)]
+#[allow(clippy::large_enum_variant)]
```
casperstorm, line 65: @tarkah Clippy think we are using too big enum now. I ignored it. I think it's because we have `Addon` inside. You think it's gonna be a problem?

tarkah: I think it's fine until we finally get around to refactoring the entire GUI module where we can figure out how to better break everything up.

PR #256, opened 2020-10-16
always refresh folders on update or install

tarkah:
This fixes a couple potential bugs... If an update to an addon has a different list of `AddonFolder` than the currently installed addon, it used to leave those old folders on the filesystem. Then the new fallback logic would show those orphaned folders as "Uknown". We now delete all top level folders based on the `Addon`s `folders` prior to extracting the new addon folders.

We also make sure that these new addon folders are parsed and assigned to the addon not only on Catalog Install, but on Update as well. This should always keep the folders of the Addon accurate.

## Checklist


tarkah:
@casperstorm this is the bug I mentioned, easy fix!

tarkah:
Worth noting that the `DownloadReason` enum I created is no longer used anywhere, since I removed that restriction on the addon folders update. I can remove this enum entirely, but it might be helpful in the future to still distinguish between when we are downloading -> unpacking -> fingerprinting because of an update vs a catalog install.

casperstorm (approved):

PR #261, opened 2020-10-17
Fix/remove all folders

### LINKED ISSUE #255: Don't attempt to write into symlinked Addons!
**Describe the bug**
When you symlink an addon into the addons Folder and download an addon with the same name (i tested my own Addon), it will write into the symlink folder and replace the files at the source which is really bad if you are a developer who doesn't use source control to recover your files :P

**Expected behavior**
Delete the link and write a new folder (after warn, because it really should let you know it's getting rid of your link) so it can never write into the link source even by misclicking.

**Software involved**
 - OS: Windows 10
 - Ajour version: 0.4.1

tarkah:
Resolves #255

## Proposed Changes
  - The old logic for removing destination folders seemed to sometimes not run depending on how the zip file got extracted. This new logic guarantees that the top level addon destination folder will get deleted before we extract. This function also removes symlinks, which resolves #255.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 
 ### Fixed
 - Fixed bug where orphaned folders could exist after updating an addon if the newer version of an addon didnt't include those folders anymore. 
+- Ensure symlinks are removed in the addons folder prior to extracting an addon, so we don't write into the symlink and instead remove the link / create a new folder.
+  - This is a request from a developer who symlinks their source code into the addons folder and Ajour could accidently overwrite it
 
 ### Packaging
 - Added Forest Night theme
--- a/crates/core/src/fs/addon.rs
+++ b/crates/core/src/fs/addon.rs
@@ -3,6 +3,7 @@ use crate::{
     parse::parse_toc_path,
     Result,
 };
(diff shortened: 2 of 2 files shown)


tarkah:
@casperstorm it's worth noting that we still also need this section: https://github.com/casperstorm/ajour/blob/5074d212c68483cd709670085d910056ae745481/crates/core/src/fs/addon.rs#L33-L36

This one ensures all old directories are cleaned up, while this new one ensures all top level addon directories that match the archive are removed (99% of the time, they are the same). I validated in the docs that the `remove_dir_all` function removes symlinks.

tarkah:
Ahh, I had to scrap what I created. It still doesn't work. To expand on the issue some more, 0.4.1 currently has this logic to delete the destination directories: https://github.com/casperstorm/ajour/blob/90481043387df23ac062e9b7aa9486f457cfc02f/crates/core/src/fs/addon.rs#L33-L39

The problem here is that we are relying on `let path = to_directory.join(file.sanitized_name());` which comes from the `ZipArchive` `File`. It seems not all archives emit the top level directory as a `File` which is why this was missed. A good example is the Recount addon. My new logic also falls prey to this gap.... Back to the drawing board....

tarkah:
Ok, that latest commit should do the trick! I get all unique top level folder names and then remove those prior to starting our extract.

PR #271, opened 2020-10-19
feat: date released column

### LINKED ISSUE #210: Catalog Sort by Date Added
Would it be possible to add a sort by date added? I love to browse the addons sections of Curse and Wowinterface to see what addons have been added or updated. Would be nice to do that from the addon.

casperstorm:
Resolves #210 

## Proposed Changes
  - This PR will add a date released column in both My Addons and Catalog.

## Checklist

### DIFF
--- a/crates/core/src/addon.rs
+++ b/crates/core/src/addon.rs
@@ -15,8 +15,8 @@ pub enum AddonVersionKey {
 pub struct RemotePackage {
     pub version: String,
     pub download_url: String,
-    pub date_time: Option<DateTime<Utc>>,
     pub file_id: Option<i64>,
+    pub date_time: Option<DateTime<Utc>>,
 }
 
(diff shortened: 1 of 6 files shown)


crates/core/src/addon.rs
```diff
@@ -258,9 +258,12 @@ impl Addon {
             let version = package.version.clone();
             let download_url = package.url.clone();
 
-            let date_time = NaiveDateTime::parse_from_str(&package.lastupdate, "%Y-%m-%d")
+            let date_time = NaiveDateTime::parse_from_str(&package.lastupdate, "%Y-%m-%d %H:%M:%S")
                 .map_or(
```
huntwj, line 262: Does it make sense to use `Result::map_or_else` here to avoid parsing a second time if it succeeds the first time. My understanding is this will parse twice every time, even if the first is successful.

PR #286, opened 2020-10-22
chore: unwrap could cause ajour to crash

casperstorm:
## Proposed Changes
  - In a rare case we assumed a little too much, and was met with a crash. I am playing a bit more defensive now and do not assume the data from CurseForge is good.

### DIFF
--- a/crates/core/src/addon.rs
+++ b/crates/core/src/addon.rs
@@ -480,14 +480,13 @@ impl Addon {
                 && info.file.modules.iter().any(|m| m.foldername == f.id)
         }) {
             f.id.clone()
+        } else if let Some(f) = addon_folders
+            .iter()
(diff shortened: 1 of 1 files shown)


crates/core/src/addon.rs
```diff
@@ -480,14 +480,13 @@ impl Addon {
...
+            f.id.clone()
         } else {
-            addon_folders
-                .iter()
-                .find(|f| info.file.modules.iter().any(|m| m.foldername == f.id))
-                .as_ref()
-                .unwrap()
-                .id
-                .clone()
+            info.file.file_name.clone()
```
casperstorm, line 489: Backlog for you @tarkah.
We had a crash here. Previously you did: 

```rs
.find(|f| info.file.modules.iter().any(|m| m.foldername == f.id)).as_ref().unwrap()
```

Apparently that is not always valid. Who could have known.

PR #290, opened 2020-10-22
Fix/better unknown support

casperstorm:
Better support for unknown/modified addons.

### DIFF
--- a/crates/core/src/addon.rs
+++ b/crates/core/src/addon.rs
@@ -112,6 +112,26 @@ pub struct AddonFolder {
     pub fingerprint: Option<u32>,
 }
 
+impl PartialEq for AddonFolder {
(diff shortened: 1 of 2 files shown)


crates/core/src/addon.rs
```diff
@@ -344,6 +365,13 @@ impl Addon {
                     file_id: Some(file.id),
                 };
 
+                let file_folders: Vec<AddonFolder> = addon_folders
+                    .iter()
+                    .filter(|f| file.modules.iter().any(|m| m.foldername == f.id))
+                    .cloned()
+                    .collect();
+                folders.extend(file_folders);
+
```
casperstorm, line 374: I run this on every file, instead of only the "main" file.
This way we ensure we get all the dependencies. Late `dedup` it.

PR #295, opened 2020-10-24
fix bug creating addons from already matched curse ids

tarkah:
Resolves issue with Altoholic brought up in Discord

## Proposed Changes
  - Any folder with a curse id that is already successfully mapped via an exact fingerprint match shouldn't allow a new addon to be created from that curse id. I've fixed a logic gap that allowed this to happen previously.

## Checklist


tarkah:
@mlablah This fixes the issue brought up this morning in discord for Altoholic

casperstorm (approved):
Nice! You can merge once you are ready.

PR #296, opened 2020-10-24
feat: identify wowinterface addons 

### LINKED ISSUE #224: Add WoWInterface as an addon source
_**Is your feature request related to a problem? Please describe.**_
I have to update addons that are exclusive to WoWInterface manually.

_**Describe the solution you'd like**_
If possible I would like to have WoWInterface included as an addon source.

casperstorm:
Resolves #224

## Proposed Changes
  - Add WoWInterface as a source.
  - Add "Source" column as a new column to see where an addon is matched towards.
  - Removed "Unknown" addons from Status column. and moved it to Source column.

## Checklist

### DIFF
--- a/crates/core/src/parse.rs
+++ b/crates/core/src/parse.rs
@@ -477,25 +477,34 @@ pub async fn read_addon_directory<P: AsRef<Path>>(
         );
     }
 
-    // Filters the WowI ids.
-    let wowi_ids: Vec<_> = addon_folders
-        .iter()
-        .filter_map(|folder| {
-            if let Some(wowi_id) = folder.repository_identifiers.wowi.clone() {
-                Some(wowi_id)
-            } else {
-                None
-            }
-        })
-        .collect();
+    let mut wowi_ids = vec![];
+    // Filter wowi addons. We only support Retail since the API is limited to that.
+    if flavor.base_flavor() == Flavor::Retail {
+        wowi_ids = addon_folders
+            .iter()
+            .filter_map(|folder| {
+                if let (Some(wowi_id), None, None) = (
+                    folder.repository_identifiers.wowi.clone(),
+                    folder.repository_identifiers.curse.clone(),
+                    folder.repository_identifiers.tukui.clone(),
+                ) {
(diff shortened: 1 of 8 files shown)


crates/core/src/parse.rs
```diff
@@ -480,8 +479,47 @@ pub async fn read_addon_directory<P: AsRef<Path>>(
...
+    let mut wowi_ids_from_nonmatch: Vec<_> = addon_folders
+        .iter()
+        .filter(|f| {
+            !fingerprint_addons.iter().any(|fa| {
+                fa.folders
+                    .iter()
+                    .any(|faf| faf.fingerprint == f.fingerprint)
+            })
+        })
+        .filter(|f| f.repository_identifiers.wowi.is_some())
```
tarkah, line 493: Do we need to make sure that tukui and curse id are None since we already created addons off any folders with those ids by this point?

casperstorm: Good point, we properly do.

casperstorm: Fixed now.

tarkah: Sweet! I still need to test but didn't notice anything else on reviewing the diff

tarkah:
@casperstorm Are we going to hold off on merging this until we implement that catalog install cache for non-curse sources?

casperstorm:
No, I think this is fine for now. But for tukui and wowi in Catalog we need it for sure.

You agree?

tarkah:
Yup, agreed!

casperstorm:
> Yup, agreed!

Will merge this to development then.

spiralofhope:
I downloaded the latest release and there doesn't appear to be any wowinterface functionality.

casperstorm:
> I downloaded the latest release and there doesn't appear to be any wowinterface functionality.

 Currently we match to against CurseForge first, if that fails (either gets no fingerprint or does not exist on CF) we look for a `wowi ` id on the addon which we can match against WoWInterface.
WoWInterface addons will also appear in the catalog from `0.5.1` :)

PR #307, opened 2020-10-26
Feat/make mac tar targets

mlablah:
Resolves #

## Proposed Changes
  - Added 2 additional targets. The workflow now also creates tar.gz assets for macos, to be used by the self-updater logic in next release.

## Checklist

Tested new assets can be run on macos.
N/A - the packaging entry for self-updater will suffice.


casperstorm:
Clean. 👍🏻

Issue #247, opened 2020-10-15
Catalog - add columns for version/game version


mattpson:
Having columns that shows remote version and/or game version in the Catalog view would be very helpful to find out that the addon you want to install is supported by the game version that you are running. 

(Patch days is horrible)

mlablah:
@mattpson thank you for the feature request. It's something we would like to add in the near future.

It has already been requested in #216 so I will close this issue. You can keep track on our progress using that other issue. :)

Issue #230, opened 2020-10-11
Catalogue filter by game version


KestrelWild:
It would be nice if there was a filter in the Catalog to only show addons that support latest game version, and then if you could still sort the filtered results by number of downloads, date updated, etc.

PR #313, opened 2020-10-26
feat: Game Version column in catalog

### LINKED ISSUE #216: Catalog columns
some addons on curse are not no longer maintained, so downloading them will slow/lag and some time crash the game, not to mention they wont work.

Adding more columns in the catalog window, the same way it is in my addons window. this way if  an addon is no longer maintained i can see that from the addon game version.

casperstorm:
Resolves #216 #247 #230 

## Proposed Changes
  - Add Game Version as a column to Catalog.
  - Optional columns in Catalog.

## Checklist

### DIFF
--- a/src/gui/mod.rs
+++ b/src/gui/mod.rs
@@ -82,8 +82,11 @@ pub enum Interaction {
     ScaleDown,
     Backup,
     ToggleColumn(bool, ColumnKey),
+    ToggleCatalogColumn(bool, CatalogColumnKey),
     MoveColumnLeft(ColumnKey),
     MoveColumnRight(ColumnKey),
+    MoveCatalogColumnLeft(CatalogColumnKey),
+    MoveCatalogColumnRight(CatalogColumnKey),
     ModeSelected(AjourMode),
     CatalogQuery(String),
     CatalogInstall(catalog::Source, Flavor, u32),
@@ -146,6 +149,7 @@ pub struct Ajour {
     scale_state: ScaleState,
     backup_state: BackupState,
     column_settings: ColumnSettings,
+    catalog_column_settings: CatalogColumnSettings,
     onboarding_directory_btn_state: button::State,
     catalog: Option<Catalog>,
     catalog_install_statuses: Vec<(Flavor, u32, CatalogInstallStatus)>,
@@ -190,6 +194,7 @@ impl Default for Ajour {
             scale_state: Default::default(),
             backup_state: Default::default(),
             column_settings: Default::default(),
+            catalog_column_settings: Default::default(),
             onboarding_directory_btn_state: Default::default(),
             catalog: None,
             catalog_install_statuses: vec![],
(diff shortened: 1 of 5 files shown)


casperstorm:
@tarkah if you have 5min at somepoint, can you run this branch and see what the hell is up with the resizing of the rows in catalog? I think I must doing something wrong. No rush on it, caching is more important.

src/gui/mod.rs
```diff
@@ -764,7 +772,7 @@ impl Default for HeaderState {
                 ColumnState {
                     key: ColumnKey::LocalVersion,
                     btn_state: Default::default(),
-                    width: Length::Units(150),
+                    width: Length::Fill,
```
tarkah, line 775: Why are we setting this as `Fill`? Shouldn't we only have one as Fill, everything else is either the default value or whatever value is the result after a resize operation?

src/gui/mod.rs
```diff
@@ -832,7 +840,7 @@ pub struct ColumnState {
 
 impl From<&ColumnState> for ColumnConfigV2 {
     fn from(column: &ColumnState) -> Self {
-        // Only `ColumnKey::Title` should be saved as Length::Fill -> width: None
+        // Only `ColumnKey::Description` should be saved as Length::Fill -> width: None
```
tarkah, line 843: It's still `ColumnKey::Title`?

src/gui/update.rs
```diff
@@ -145,8 +193,29 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
                         }
                     });
```
tarkah, line 195: We're missing the `ajour.catalog_header_state.columns.iter()` here, which is probably messing things up

tarkah:
@casperstorm caught the couple tiny issues that were holding it up. Took me a while to catch because everything starts looking the same 😵 Great stuff!

src/gui/update.rs
```diff
@@ -85,25 +85,6 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
...
-                                    None
-                                }
-                            })
-                            .next()
-                        {
-                            a.width = column.width.map_or(Length::Fill, Length::Units);
-                            a.hidden = column.hidden;
-                            a.order = idx;
-                        }
-                    });
```
tarkah: `ColumnConfig::V2` is from before we implemented the catalog, so this config only relates to the MyAddons stuff

casperstorm:
> @casperstorm caught the couple tiny issues that were holding it up. Took me a while to catch because everything starts looking the same 😵 Great stuff!

Thanks for the assistance @tarkah and yeah its hard to get lost with this naming. Maybe we could consider moving it to a seperate file as well, eventually.

casperstorm:
@tarkah could you give it one more look on the review side? I think im done now :)

PR #315, opened 2020-10-27
feat: Addon Cache

tarkah:
Resolves adding Tukui and WowI reliably to catalog

## Proposed Changes
  - Cache addon api and folder grouping data for Tukui and WowI addons installed through catalog since those APIs don't give us fingerprinting / modules like Curse and the addon folders are unreliable for getting repo IDs and impossible to determine grouping off of
  - Use this cache when parsing addons so we can reliably group / get repo metadata for these addons

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -82,6 +82,7 @@ dependencies = [
  "fancy-regex",
  "fern",
  "flate2",
+ "futures",
  "glob",
  "iced_native",
  "isahc",
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -41,6 +41,7 @@ log = "0.4"
 fern = "0.6"
 walkdir = "2.3"
 retry = "1.1"
+futures = "0.3"
 
 iced_native = { git = "https://github.com/hecrj/iced.git", rev = "d3b04bf892ce63d3129686968039c258945c5b02", optional = true }
 
--- a/crates/core/src/addon.rs
+++ b/crates/core/src/addon.rs
@@ -284,12 +284,17 @@ impl Addon {
         }
     }
 
-    /// Creates an `Addon` from the WowI package
-    pub fn from_wowi_package(
+    /// Updates an `Addon` from the WowI package.
+    ///
+    /// When addon_folders is passed, it will update the primary folder id and mapped
+    /// addon folders. This is only used when creating the addon via fallback measure.
+    /// WowI addons cached will have these updated from the cache.
+    pub fn update_with_wowi_package(
+        &mut self,
         wowi_id: i64,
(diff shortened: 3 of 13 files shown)


tarkah:
Just need to account for the cache during parsing and this will be done!

mlablah:
I was sent a build from this branch and I had a go using it. I pointed it at a directory where I have top 50 retail addons from curseforge (listed as 69 addons due to altoholic containing other addons), installed via catalog 10 days ago. 

There are also addons in the folders for the other game flavors but those can be ignored. 

I then hit "Update All" and ajour panicked and shut down. Attached is zipped version of ajour's appdata and a screenshot showing  14 temp files left behind afterwards. 

My guess is that the panic was caused by altoholic which has matched addons that are subsets of another matched addon causing problems when altoholic and its sub-addons are trying to update the same files. I don't know if the problem is new to this branch.

08:43:16.230 [panic][ERROR] thread 'main' panicked at 'called `Option::unwrap()` on a `None` value': src\gui\update.rs:788

[logs.zip](https://github.com/casperstorm/ajour/files/5456877/logs.zip)

casperstorm:
Looks like its  DataStore_Reputations (`2623891795`) which causes the crash. It couldn't create a `primary_folder_id` from the response.

```rust
folders.get(0).map(|f| f.id.clone()).unwrap()
```

tarkah:
Yeah, so what appears to be happening is that due to how FUBAR Altoholic is, it's trying to download the main Altoholic addon, on top of the individual addon for each Datastore. Because the main Altoholic addon contains each folder, there is a high chance that there will be a collision while it's unpacking one of the Datastore folders, with the unpacking on that same Datastore folder from the individual addon. At the end of our unpack function, we parse all the unpacked folders into `AddonFolder`

```rust
let addon_folders = toc_files.iter().filter_map(parse_toc_path).collect();
```

When it was doing this for the individual addon for ` DataStore_Reputations`, the main Altoholic addon was also unpacking, and was probably mid-unpacking it's ` DataStore_Reputations` folder, while the other one was trying to parse it. When it tried to parse it, the `.toc` didn't exist (due to the collision, it was still being fully unpacked) and therefore it returned an empty Vec. Which is why this failed:

```rust
//TODO: Can this crash? What do we do in that case.
folders.get(0).map(|f| f.id.clone()).unwrap()
```

Because the were no successfully parsed AddonFolder in the vec.

I've added `if !folders.is_empty() {` as protection in this scenario.

casperstorm:
No remarks. Looks good, and tests was good.

tarkah:
@casperstorm just double check my changes to config loading initially and I think this is good to go

PR #317, opened 2020-10-29
Fix/linux size scale

### LINKED ISSUE #190: Window size issues with scaling on Linux
**Describe the bug**
I noticed that when I have scaling applied (for example, 1.20) and I close the application and then re-open, the window size increases by that 20%. Same goes when scale is < 1.0, window size shrinks each time.

**To Reproduce**
1. On linux, increase / decrease scale
2. Close window
3. Reopen window
4. Repeat

**Expected behavior**
Window size is consistent between sessions

**Software involved**
 - OS: Linux (Solus OS)
 - Ajour version: 0.3.5

**Additional context**
Totally caused by this, which was a workaround to get things playing nicely on Windows, but obviously has different behaviour on linux. Though it's unknown if its just a problem on Gnome based DE or Plasma as well... Does this happen on mac?

https://github.com/casperstorm/ajour/blob/442e9b9963d5b94315d5170ec4a2c9393b1379de/src/gui/update.rs#L913-L917

I can fix this w/ conditional compilation, we just need to test on each platform / multiple linux DEs to make sure we understand where it's happening.

tarkah:
Resolves #190 

## Proposed Changes
  - We pass the scale as a Application flag so our scale is set on the very first loop of our program. That way when the "rogue" resize event on linux is received, the dimensions aren't incorrectly changed by our scale factor adjustment logic. The problem was the "rogue" resize event happened before the config got loaded / scale was updated, so the resize event width / height weren't "scale adjusted" and then our scale adj logic fucked it up and saved it to the config, which then bit you in the ass on next launch.
  - I've also fixed the issue with saving floating point arithmetic unrounded numbers of the scale by forcing the f64 to round to 1 decimal place.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 - Cancelling when changing wow path will empty list.
 - Fixed case-sensitive issue when sorting addons by title. 
 - Better addon changelog formatting.
+- Fixed bug on linux that caused window size to grow / shrink between sessions when a <>1.0 scale was set
 
 ### Packaging
 - Ajour can now self update when a new release is available. An "Update" button will appear along with a message that a newer release is available. Clicking this button will automatically update Ajour and relaunch it as the newer version.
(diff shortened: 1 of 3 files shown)


tarkah:
@casperstorm I'm so happy to have found a fix for this bug on linux (not sure if it affected ALL distros, but definitely did with the Budgie Desktop Environment). I need to have a scale applied on my hidpi display, but if I did (I prefer 1.3), the window grew 30% every time I relaunched it!! Now it stays put :)

casperstorm:
Works, and great we solved this.

PR #319, opened 2020-10-29
refactor GUI

tarkah:
## Proposed Changes
  - Refactor gui so our eyes stop bleeding after looking at it
  - Split up the layout into "components" and push the state, update & view logic into those components
  - Layout components across modules / filesystem so each module has its own component definition (no more 1k+ lines per module)

## Checklist


tarkah:
c4c5910d12d8046039a5977c0b193b635ba5f234 has a mockup of my proposed "Component" trait and a test component for how it'd be used

valms:
Hello! How can i help you guys?

PR #320, opened 2020-10-29
chore: refactor state & mode

casperstorm:
Resolves multiple problems related to the fact that we had a single `AjourState` enum controlling loading state for multiple areas in the app.

## Proposed Changes
  - Idea is to split it into a `HashMap<Mode, State>` so each mode have its own state.
  - `Mode:MyAddon` now takes `flavor` we each `flavor` also has a `State`.
  - Error is removed from `State` and is now living on the `Ajour` enum.
  
I am missing a few things still, one thing I would like to tackle is how to properly handling multiple flavors loading.

## Checklist

### DIFF
--- a/src/gui/element.rs
+++ b/src/gui/element.rs
@@ -2,10 +2,10 @@
 
 use {
     super::{
-        style, AddonVersionKey, AjourMode, AjourState, BackupState, CatalogColumnKey,
-        CatalogColumnSettings, CatalogColumnState, CatalogInstallStatus, CatalogRow, Changelog,
-        ColumnKey, ColumnSettings, ColumnState, DirectoryType, ExpandType, Interaction, Message,
-        ReleaseChannel, ScaleState, SelfUpdateState, SortDirection, ThemeState,
+        style, AddonVersionKey, BackupState, CatalogColumnKey, CatalogColumnSettings,
+        CatalogColumnState, CatalogInstallStatus, CatalogRow, Changelog, ColumnKey, ColumnSettings,
+        ColumnState, DirectoryType, ExpandType, Interaction, Message, Mode, ReleaseChannel,
+        ScaleState, SelfUpdateState, SortDirection, State, ThemeState,
     },
     crate::VERSION,
     ajour_core::{
@@ -20,6 +20,7 @@ use {
         HorizontalAlignment, Length, PickList, Row, Scrollable, Space, Text, VerticalAlignment,
     },
     num_format::{Locale, ToFormattedString},
+    std::collections::HashMap,
     widgets::{header, Header},
 };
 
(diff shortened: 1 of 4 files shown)


src/gui/mod.rs
```diff
@@ -160,6 +166,9 @@ pub struct Ajour {
 impl Default for Ajour {
     fn default() -> Self {
         Self {
+            state: [(Mode::Catalog, State::Loading)].iter().cloned().collect(),
```
casperstorm, line 169: Initially start catalog in `State::Loading` state as this is the only part we know for sure will be there.
The `MyAddon(flavor)` will be added later.

src/gui/mod.rs
```diff
@@ -315,7 +323,7 @@ impl Application for Ajour {
         content = content.push(Space::new(Length::Units(0), Length::Units(DEFAULT_PADDING)));
 
         match self.mode {
-            AjourMode::MyAddons => {
+            Mode::MyAddons(flavor) => {
```
casperstorm, line 326: We now have flavor from the mode we have selected. A bit prettier than asking `config.wow.flavor` very single time.

casperstorm:
Give this a run, @tarkah.

src/gui/element.rs
```diff
@@ -1386,6 +1394,12 @@ pub fn menu_container<'a>(
     classic_ptr_btn_state: &'a mut button::State,
     self_update_state: &'a mut SelfUpdateState,
 ) -> Container<'a, Message> {
+    let flavor = config.wow.flavor;
+
+    // State.
+    let default = &State::default();
+    let myaddons_state = state.get(&Mode::MyAddons(flavor)).unwrap_or(default);
```
tarkah, line 1401: I think this feels more idiomatic, since we can used cloned to get a Option<State> then use unwrap_or_default. It would also be worthwhile adding derive(Copy) on State since its sized. We should always try to derive Copy along w/ Clone where we can

```diff
-    let default = &State::default();
-    let state = state.get(&Mode::MyAddons(flavor)).unwrap_or(default);
+    let state = state
+        .get(&Mode::MyAddons(flavor))
+        .cloned()
+        .unwrap_or_default();
```

src/gui/mod.rs
tarkah: Makes sense!

src/gui/update.rs
```diff
@@ -372,8 +380,13 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
                 // Persist the newly updated config.
                 let _ = &ajour.config.save();
                 // Set loading state.
-                ajour.state = AjourState::Loading;
-                // Reload config.
+                let state = ajour.state.clone();
```
tarkah, line 383: With derive(Copy) on state, we won't need to explicitly clone it

tarkah (approved):

PR #323, opened 2020-10-31
chore: update title regex

### LINKED ISSUE #326: TukUI addon using title from toc
**Describe the bug**
I am using Ajour 0.5.0 from the Development branch.

This is a combination of two issues. One revealing the other.

A tukui addon called ElvUI_AbsorbTags appears with web color codes in the title after being installed. 

`ElvUI cff83F3F7Absorb Tags`

From the toc file it looks like this: `## Title: |cff1784d1ElvUI |cff83F3F7Absorb`

So it appears as though the first web color tag has been removed but the second hasn't. Possibly due to its use of mixed upper/lower case hexadecimal characters.

However we should only use the title from the toc file as a fallback for unrecognized addons. So the bigger issue is why is it even using the title from the toc file and not the name from the api?

Here is the api results for that addon: https://www.tukui.org/api.php?addon=85
And here is the title as it appears on the website: https://www.tukui.org/addons.php?id=85

Notice both have a name with an underscore and no color codes.

From the cache (addons.yml), the addon looks like this:
```
V1:
  Retail:
    - title: ElvUI cff83F3F7Absorb Tags
      repository: Tukui
      repository_id: "85"
      primary_folder_id: ElvUI_AbsorbTags
      folder_names:
        - ElvUI_AbsorbTags
      modified: "2020-10-31T23:00:45.692779100Z"
```

So it seems we save the (partially cleaned) toc title instead of the name from the api, as the title.

**To Reproduce**
1. Go to the Catalog
2. Search for and install the retail addon ElvUI_AbsorbTags using TukUI source.
3. Go to My Addons and see the title contains web color code.

**Expected behavior**
That the addon was using title as listed in api response. Or failing that, using a fully cleaned version of title from toc.

**Screenshots**
unknown

**Software involved**
 - OS: Windows 10
 - Ajour version: 0.5.0 from development branch.

**Log Output**
[ajour.log](https://github.com/casperstorm/ajour/files/5470751/ajour.log)

casperstorm:
Partial fixes #326.

## Proposed Changes
- Updated the regex for parsing toc titles.
- Refactored so we now use `lazy_static` to avoid compiling the same regexes multiple times
- Added tests

### DIFF
--- a/crates/core/src/parse.rs
+++ b/crates/core/src/parse.rs
@@ -1101,7 +1101,7 @@ pub fn parse_toc_path(toc_path: &PathBuf) -> Option<AddonFolder> {
 
     // TODO: We should save these somewere so we don't keep creating them.
     let re_toc = regex::Regex::new(r"^##\s*(?P<key>.*?)\s*:\s?(?P<value>.*)").unwrap();
-    let re_title = regex::Regex::new(r"\|[a-fA-F\d]{9}([^|]+)\|?r?|\.?\|T[^|]*\|t").unwrap();
+    let re_title = regex::Regex::new(r"\|(?:[a-fA-F\d]{9}|T[^|]*\|?|t|r)").unwrap();
 
     for line in reader.lines().filter_map(|l| l.ok()) {
         for cap in re_toc.captures_iter(line.as_str()) {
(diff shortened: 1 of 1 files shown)


casperstorm:
I'll make some tests to this before merging.

PR #330, opened 2020-11-01
fix catalog install flow

tarkah:
## Proposed Changes
- Do not insert the catalog addon into our main addons vec until it has
  been successfully unpacked.
- Make sure any existing addon that has the same addon folders as our
  newly installed addon is removed prior to inserting the new addon to
  prevent dupliates.
- Make sure addon cache entries are removed when the addon is deleted
  or when the same addon is installed via catalog from the Curse source

## Checklist

### DIFF
--- a/crates/core/src/addon.rs
+++ b/crates/core/src/addon.rs
@@ -870,6 +870,39 @@ impl Addon {
             }
         }
     }
+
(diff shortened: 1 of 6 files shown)


casperstorm (approved):
Looks good 👍🏻

casperstorm:
You feel this is ready to be merged, @tarkah?

tarkah:
Yup!

PR #335, opened 2020-11-03
feat: wowi addons in catalog

casperstorm:
## Proposed Changes
  - Adds WowInterface addons to the catalog.

## Checklist

- [x] Merge https://github.com/casperstorm/ajour-catalog/pull/9
- [x] Revert to `catalog.json`
- [x] Update to release version of catalog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 
 ## [Unreleased]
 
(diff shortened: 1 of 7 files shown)


casperstorm:
@tarkah this should be done, wanna take a look?

tarkah (approved):
Looks good! Super straight forward. 

I also did a quick check in all places we used `Repository` enum and it looks like WowI was already being handled everywhere perfectly, so that all looks good too!

PR #356, opened 2020-11-15
feat: reworked settings and added about

casperstorm:
## Proposed Changes
  - Reworking Settings so its easier to maintain.
  - Added About

Screenshot 2020-11-18 at 21 33 35
Screenshot 2020-11-18 at 21 33 38

## Checklist


tarkah:
@casperstorm Just tried out this PR, looking good! I added some logic if `ESC` is pressed within `About` or `Settings` mode, it'll revert to `MyAddons`. Also if the `About` or `Settings` buttons are clicked when those modes are already active, it'll toggle off them back to `MyAddons` as well. I'd still consider an explicit "Close" button somewhere as it might not be obvious that the user needs to explicitly toggle off these 2 modes or select MyAddons to get off it.

Also, I'd prefer to see some more left / top padding in the Setting view, the contents are too close to the edges for my taste.

casperstorm:
> @casperstorm Just tried out this PR, looking good! I added some logic if `ESC` is pressed within `About` or `Settings` mode, it'll revert to `MyAddons`. Also if the `About` or `Settings` buttons are clicked when those modes are already active, it'll toggle off them back to `MyAddons` as well. I'd still consider an explicit "Close" button somewhere as it might not be obvious that the user needs to toggle off these 2 modes via it's button or press MyAddons/Catalog/etc to get off it.
> 
> Also, I'd prefer to see some more left / top padding in the Setting view, the contents are too close to the edges for my taste.

Cool, I like that. Good job.
I've added some more padding as well. Should be good now.

casperstorm:
@tarkah take a look at this now :)

PR #358, opened 2020-11-16
Fetch and parse catalog sources asynchronously

Mazwak:
Resolves #

## Proposed Changes
  - Parse JSON catalogs in a thread
  - - It resolves a timeout on another async req
  - - It makes use of multi cpu

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 
 ## Added
 - Option to hide ignored addons.
+- Load API catalogs asynchronously.
 
 ### Fixed
 - Game Version fallback to TOC could in some cases fail
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,10 +11,10 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 
 ## Added
 - Option to hide ignored addons.
-- Load API catalogs asynchronously.
 
 ### Fixed
 - Game Version fallback to TOC could in some cases fail
+- Load API catalogs asynchronously.
 
 ## [0.5.1] - 2020-11-12
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 
 ### Fixed
 - Game Version fallback to TOC could in some cases fail
-- Load API catalogs asynchronously.
+- Load addons into catalog asynchronously.
 
 ## [0.5.1] - 2020-11-12
 
--- a/crates/core/src/catalog.rs
+++ b/crates/core/src/catalog.rs
@@ -1,9 +1,9 @@
 use crate::config::Flavor;
-#[macro_use]
 use crate::error;
 use crate::Result;
-use async_std::{fs::File, io, prelude::*, task};
+use async_std::task;
 use chrono::prelude::*;
+use futures::future::join_all;
 use std::time::Duration;
 
 use isahc::{config::RedirectPolicy, prelude::*};
@@ -43,15 +43,17 @@ pub async fn get_one_catalog(url: &str) -> Result<Vec<CatalogAddon>> {
 pub async fn get_catalog() -> Result<Catalog> {
     let mut handles = vec![];
     for url in CATALOG_URLS.iter() {
-        handles.push(task::spawn(async { get_one_catalog(url) }));
+        let url = url.clone();
+        handles.push(task::spawn(async move { get_one_catalog(url) }));
     }
 
     let mut addons = vec![];
-    for h in handles {
-        let catalog = h.await;
-        match catalog {
-            Ok(mut c) => addons.append(&mut c.addons),
-            Err(e) => println!("TODO"),
+    let results = join_all(handles).await;
+    for api_results in results {
+        let r = api_results.await;
+        match r {
+            Ok(c) => addons.append(&mut c.clone()),
+            Err(_) => log::debug!("Could not get one catalog"),
         };
     }
 
--- a/crates/core/src/catalog.rs
+++ b/crates/core/src/catalog.rs
@@ -30,7 +30,6 @@ pub async fn get_catalog_addons_from(url: &str) -> Result<Vec<CatalogAddon>> {
 
     let request = client.get_async(url.to_string());
     if let Ok(mut response) = request.await {
-        log::debug!("Fetched {}, beginning parsing", url);
         if let Ok(json) = task::spawn_blocking(move || response.json()).await {
             log::debug!("Successfully fetched and parsed {}", url);
             Ok(json)
--- a/crates/core/src/catalog.rs
+++ b/crates/core/src/catalog.rs
@@ -46,7 +46,6 @@ pub async fn get_one_catalog(url: &str) -> Result<Vec<CatalogAddon>> {
 pub async fn get_catalog() -> Result<Catalog> {
     let mut handles = vec![];
     for url in CATALOG_URLS.iter() {
-        let url = url.clone();
         handles.push(task::spawn(async move { get_one_catalog(url) }));
     }
 
--- a/crates/core/src/catalog.rs
+++ b/crates/core/src/catalog.rs
@@ -1,10 +1,14 @@
 use crate::config::Flavor;
-use crate::network::request_async;
 use crate::Result;
 use chrono::prelude::*;
+use std::sync::mpsc::channel;
+use std::time::Duration;
 
 use isahc::{config::RedirectPolicy, prelude::*};
 use serde::Deserialize;
+use threadpool::ThreadPool;
+
+const DEFAULT_TIMEOUT: u64 = 30;
 
 const CURSE_CATALOG_URL: &str =
     "https://github.com/casperstorm/ajour-catalog/releases/latest/download/curse.json";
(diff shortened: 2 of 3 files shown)


tarkah:
@Mazwak to add some more context to our earlier conversation...

`async_std` already runs it's own thread pool, by default it spawns one thread for each logical core on the CPU, see: https://docs.rs/async-std/1.7.0/async_std/index.html#runtime-configuration. So having 3 futures, one for processing catalog file, should be distributed on the threadpool. We just need to make sure any blocking operations get wrapped in https://docs.rs/async-std/1.7.0/async_std/task/fn.spawn_blocking.html (our deserialization).

We are still requesting the catalog in sequence here. The code we write is still resolved normally, and the `await` keyword signifies we are waiting on the background task to complete before moving forward. So we are blocking this on each api request.
https://github.com/casperstorm/ajour/blob/2962099133150b732612c33bc46a68d1ce86e222/crates/core/src/catalog.rs#L32-L34

My suggestion would be to create a Future that does the following:

- Request json from API
- spawn_blocking the deserialization step
- returns Vec<CatalogAddon>

We can create all 3 futures, then spawn them to be worked on concurrently with https://docs.rs/futures/0.3.8/futures/future/fn.join_all.html and collect the result into a single vec... something like `join_all(the_3_futures).await.iter().flatten().collect()`

tarkah:
You can see an example of using `join_all` in the code here. `.map(|id| tukui::fetch_remote_package(client.clone(), &id, &flavor)) ` returns a Future, so we basically have a `Vec<Future>` which we then pass into `join_all`.

You can create an async function to return the future, like we are doing here, or you can even just return an async block inside the `map` like `.map(|_| async {})` and define the steps I listed above inside that async block.

https://github.com/casperstorm/ajour/blob/96d3edc71f3bdcb2130e91cfcd30e4c1545c269b/crates/core/src/parse.rs#L528-L535

Mazwak:
On « spawn_blocking », I’m not sure it is still mandatory.
See : https://async.rs/blog/stop-worrying-about-blocking-the-new-async-std-runtime/

I didn’t found if it was stable yet or not.

Mazwak:
I’m stuck with this error:

error[E0308]: mismatched types
  --> crates\core\src\catalog.rs:54:13
   |
24 | pub async fn get_one_catalog(url: &str) -> Result<Vec<CatalogAddon>> {
   |                                            ------------------------- the `Output` of this `async fn`'s expected opaque type
...
54 |             Ok(mut c) => addons.append(&mut c.addons),
   |             ^^^^^^^^^ expected opaque type, found enum `std::result::Result`
   |
   = note: expected opaque type `impl futures::Future`
                     found enum `std::result::Result<_, _>`

I more or less understant why, but not how to correct it

Mazwak:
It works now.

But I don’t understand why I need to await twice.

Mazwak:
Now, it does run asynchronously.

But it feels bad to do « spawn_blocking » on « block_on ».

Mazwak:
Should be close to final now

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 
 ## Added
 - Option to hide ignored addons.
+- Load API catalogs asynchronously.
```
casperstorm, line 14: Let's move this to `## Changed`or `## Fixed`.

Mazwak: Done

Mazwak:
Tested OK on windows.

After cargo update, both debug and release.

launched. Added an addon. Deleted it.

crates/core/src/catalog.rs
```diff
@@ -1,43 +1,63 @@
...
     }
+}
 
-    if tukui_resp.status().is_success() {
-        let mut catalog: Catalog = tukui_resp.json()?;
-        addons.append(&mut catalog.addons);
+pub async fn get_catalog() -> Result<Catalog> {
+    let mut futures = vec![];
+    for url in CATALOG_URLS.iter() {
+        // handles.push(task::spawn_blocking(move || { task::block_on(get_one_catalog(url)) }));
```
casperstorm, line 50: Should remove this.

Mazwak: Done

Mazwak:
I’m also wondering if « get_one_catalog » should just return a Vec<CatalogAddon>>, without Result<>.

Just logging and returning an empty Vec on error.

casperstorm:
> I’m also wondering if « get_one_catalog » should just return a Vec>, without Result<>.
> 
> Just logging and returning an empty Vec on error.

Now that you open up the nitpicking conversation heres my 2 cent:
I think it is obvious by the argument that it is `1` catalog, so we could remove that to just `get_catalog(&str)`.
I think its fine it returns a result so it does nothing else than that. In these situations I would assume it's a `Result` because I know it can fail.

Mazwak:
I used `get_one_catalog` because `get_catalog` is already used.

I think it should be get_catalog and get_catalogs. Which means refactoring all the places where it is called. Should only be one or two though.

casperstorm:
> I used `get_one_catalog` because `get_catalog` is already used.
> 
> I think it should be `get_catalog` and `get_catalogs`. Which means refactoring all the places where it is called. Should only be one or two though.

Hmm true. Well, we are always talking about `one catalog` (with multiple sources). So maybe lets stick with the public function to be `get_catalog` and then we could have `get_catalog_source` or something like that.

Mazwak:
The best I can think of is `fetch_addons_from`.

Would make the call be : `fetch_addons_from(url)

casperstorm:
> The best I can think of is `fetch_addons_from`.
> 
> Would make the call be : `fetch_addons_from(url)
> 
> or `get_addons_from`

`get_catalog_addons_from` maybe just to be more precise?

Mazwak:
I left a lot of log::debug!. I needed them to understand what was going on.

Should they be deleted ?

crates/core/src/catalog.rs
```diff
@@ -1,43 +1,62 @@
...
-    let mut tukui_resp = request_async(&client, TUKUI_CATALOG_URL, vec![], Some(30)).await?;
-    let mut wowi_resp = request_async(&client, WOWI_CATALOG_URL, vec![], Some(30)).await?;
-
-    let mut addons = vec![];
-    if curse_resp.status().is_success() {
-        let mut catalog: Catalog = curse_resp.json()?;
-        addons.append(&mut catalog.addons);
+    let request = client.get_async(url.to_string());
+    if let Ok(mut response) = request.await {
+        log::debug!("Fetched {}, beginning parsing", url);
```
casperstorm, line 33: I think we can remove this.

Mazwak:
I also went directly with `get_async` instead of the internal `request_async`.

It might allow reuse of http client. I think it would not change much though.

crates/core/src/catalog.rs
Mazwak: Done

casperstorm (approved):

tarkah:
@Mazwak Awesome stuff!! I'll further review & test this a bit later today and merge if all is good.

crates/core/src/catalog.rs
```diff
@@ -1,43 +1,61 @@
...
 
-    if wowi_resp.status().is_success() {
-        let mut catalog: Catalog = wowi_resp.json()?;
-        addons.append(&mut catalog.addons);
+    let mut addons = vec![];
+    let results = join_all(futures).await;
+    for api_result in results {
+        match api_result {
+            Ok(c) => addons.append(&mut c.clone()),
+            Err(e) => log::debug!("{}", e),
```
tarkah, line 57: Instead of logging the error here (it was already logged under `get_catalog_addons_from`), we should actually return it. We display errors in the GUI so the user has feedback to know something went wrong. Currently this function doesn't actually return any error.

tarkah (changes requested):
Please return the error as per the other comment, otherwise this looks good!

crates/core/src/catalog.rs
casperstorm: Good catch.

Mazwak: I didn’t want to change function signature.
This means changing calling site, since the function could only return Ok so far.

Moreover, if one source fails, it doesn’t mean that all failed.
I would return an error only if all sources fails.
The debug here is superfluous though, as it doesn’t have the exact source anyway.

There could be a notification bus which the gui could use to have a notification area maybe. But that’s a much bigger change.

casperstorm: What do you mean with ` since the function could only return Ok so far.`.
It's a `Result` so it can return a `ClientError` as well :)

Mazwak: Yes, but from the code, it is only returning Ok(). It will never return an error.

So the calling site might just ignore errors. I didn’t checked.

Mazwak: Made it so it returns an error only if all sources fails.

crates/core/src/catalog.rs
```diff
@@ -1,46 +1,72 @@
...
 
-    if wowi_resp.status().is_success() {
-        let mut catalog: Catalog = wowi_resp.json()?;
-        addons.append(&mut catalog.addons);
+    let mut addons = vec![];
+    let mut any_ok = false;
+    let results = join_all(futures).await;
+    for api_result in results {
+        match api_result {
+            Ok(c) => {
```
tarkah, line 57: Please fix the clippy lint here. We don't need to match since we aren't using the Err(_) pattern. You can change it to 

```rust
if let Ok(c) = api_result {
    addons.append(&mut c.clone());
    any_ok = true;
}
```

tarkah (changes requested):

crates/core/src/catalog.rs
tarkah: Good call

crates/core/src/catalog.rs
```diff
@@ -1,46 +1,69 @@
...
+    let results = join_all(futures).await;
+    for api_result in results {
+        if let Ok(c) = api_result {
+            addons.append(&mut c.clone());
+            any_ok = true
+        }
     }
 
-    Ok(Catalog { addons })
+    if any_ok {
```
casperstorm, line 62: Couldn't we just check if `addons.is_empty` here?

tarkah (approved):
Awesome job on your first contribution to Ajour!

crates/core/src/catalog.rs
tarkah: Good catch. @Mazwak, sorry! One more change :P

tarkah (changes requested):

crates/core/src/catalog.rs
Mazwak: They are semantically very close indeed.

The only edge case, and both variations are wrong, is if jsons are empty, for whatever reasons.

maybe, pseudo code : addons.any( |c| !c.empty)

Gonna try to code this.

tarkah: Jsons should never be empty. If they are, that means there was an error fetching them. Therefore empty addons == some error fetching.

tarkah (approved):

crates/core/src/catalog.rs
Mazwak: I was wrong anyway. Changed.

tarkah:
Merged! 🎉

crates/core/src/catalog.rs
Mazwak: I was wrong anyway. Changed.

PR #363, opened 2020-11-18
Categories per source

### LINKED ISSUE #360: Do not allow empty categories
Problem :

Categories are tied to an API (curse, wowi), but the APIs do not always share the same name.

So if you select API1 and a category from API2, the addon list ends empty, without much feedback on why.

Solution :

1) Make it so the combo only shows categories from the selected API

2) Rename categories so they are the same in both API

Mazwak:
Resolves #360

## Proposed Changes
  - Only list categories relevant to the source selected

## Checklist

- [x] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 ### Fixed
 - Game Version fallback to TOC could in some cases fail
 - Load addons into catalog asynchronously.
+- Only show categories pertaining to the source selected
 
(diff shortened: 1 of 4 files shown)


Mazwak:
It works, but there is a function to extract, as I copy/paste from calagog load into source changed

Mazwak:
This should be better now.

Lists are only processed once.
There is almost no copy paste.

casperstorm:
@tarkah any feedback? LGTM.

tarkah:
Looks good here as well. Great job @Mazwak!

PR #370, opened 2020-11-20
Catalog date order on first click

Mazwak:
Makes it so, in the catalog, if you click on 'date released', it sorts descending by default.

I believe that’s what 99.9 % of users expect.

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 - Game Version fallback to TOC could in some cases fail
 - Load addons into catalog asynchronously
 - Only show categories pertaining to the source selected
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ The sections should follow the order `Packaging`, `Added`, `Changed`, `Fixed` an
 - Game Version fallback to TOC could in some cases fail
 - Load addons into catalog asynchronously
 - Only show categories pertaining to the source selected
+- Date sort catalog descending first, which is more natural
```
casperstorm, line 22: Please move this to `Changed`.

Mazwak: Done.

Also fixed 'added'

PR #376, opened 2020-11-21
Chore/redo error handling

tarkah:
## Proposed Changes
  - Better concrete error types out of `ajour_core`
  - Adding human readable context to errors at application level and only displaying that to end users via gui
  - Better logging of errors at application level, logging both human readable context and underlying error

## Checklist

- [x] Tested on Windows
- [x] Tested on MacOS
- [x] Tested on Linux


crates/core/src/lib.rs
```diff
@@ -12,7 +12,3 @@ pub mod repository;
 #[cfg(feature = "gui")]
 pub mod theme;
 pub mod utility;
-
-use crate::error::ClientError;
-
-pub type Result<T> = std::result::Result<T, ClientError>;
```
casperstorm, line 18: Nice we got rid of this.

tarkah: yup! We have a few different error types now

casperstorm:
No more comments here, this is awesome.

casperstorm (approved):

casperstorm:
You can merge once you feel this is ready, since its a big one @tarkah.

tarkah:
Sounds good! I'll back merge dev since we've added some more stuff and get it merged later

PR #378, opened 2020-11-21
feat: adding config dir button

casperstorm:
#  Proposed Changes
  - Adding a button to open config dir in Settings
 
## Checklist

- [x] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux


casperstorm:
@tarkah Can you test this on Linux and review it?

casperstorm:
Note renamed previous `Message ` from `OpenDirectory `to `SelectDirectory` since thats what we did. Then i created a new `Message `which is called `OpenDirectory` which actually opens a directory.

PR #381, opened 2020-11-22
Fix/null values

tarkah:
Resolves issue discussed in discord

## Proposed Changes
  - Add a custom deserializer to deser null as Default. This way we don't need to  specify our fields as `Option<T>`.
  - Added a new test to validate this works

## Checklist

- [ ] Tested on Windows
- [x] Tested on MacOS
- [x] Tested on Linux

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ and `Removed`.
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


casperstorm (approved):
LGTM, Good job mate.

PR #389, opened 2020-11-25
feat: min window size

casperstorm:
## Proposed Changes
  - Added a minimum size for Ajour window.

## Checklist

- [ ] Tested on Windows
- [X] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ and `Removed`.
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


jokeyrhyme:
I think this might have caused a regression for Linux / wayland

My ajour window is now unable to be resized at all

casperstorm:
@tarkah Can you test this?

PR #393, opened 2020-11-26
chore: refactored AddonState

casperstorm:
## Proposed Changes
  - If download fails on a addon we show a  retry button.

## Checklist

- [ ] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux


tarkah (changes requested):
You can set addon state to Retry on the fail branch of Message::UnpackedAddon as well

casperstorm:
This should be good to go now @tarkah

Issue #241, opened 2020-10-14
Highlight addon row on catalog view


t03r1cht:
**Is your feature request related to a problem? Please describe.**
I find it hard to keep track which row in the catalog row my cursor is currently positioned on. When I want to download the addon I need to cautiously move my cursor over to the Retail/Classic download button.

**Describe the solution you'd like**
Visual indication (e.g. changing the color) of the row which is currently focused or let me reorder the Catalog colums like in the My Addons view so I can directly put the Downloads column next to the Addon Name column.

casperstorm:
This is a great idea, and something i've had in mind for sometime.
Problem is executing it with the current setup and ui framework we use.

I'll keep it ind mind, because UX wise it is needed.

xPucTu4:
I just came here to post about this. On wide monitors the using maximized window is a bit challenging.

casperstorm:
> Currently the header row supports cell highlight on mouse over. It should not be that hard to extend it down for each row.

Feel free to take a stab at it and send us a PR with your suggestion. It would indeed be a nice to have a good solution for this.

PR #403, opened 2020-12-01
Highlight row on mouseover

kawashita86:
Hello,
first of all thanks for developing this app, it's amazing and each upgrade is for the better!
As i found that highlighting a row was a requested feature (https://github.com/casperstorm/ajour/issues/241) I've tried to implement it. Since i'm a newbie in Rust, and just started diving into the language, I went for an easy solution that can be helpful for the eye.

Resolves #

## Proposed Changes
  -  Implemented a custom widget "TableRow" that is a superset of iced_native::container and that allows a "hovered" state.
  -  Replaced the type of element::catalog_data_cell and element::addon_data_cell from Container to TableRow
  - Added custom styling for the new type by following the guideline of the other element in gui/style.rs to show a border around  hovered row 
image

## Checklist

- [x] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/crates/widgets/src/renderer/table_row.rs
+++ b/crates/widgets/src/renderer/table_row.rs
@@ -33,9 +33,14 @@ where
             content.draw(self, &defaults, content_layout, cursor_position, viewport);
 
         (
-            if style.background.is_some() || style.border_width > 0.0 {
+            if is_mouse_over && (style.background.is_some() || style.border_width > 0.0) {
                 let background = Primitive::Quad {
-                    bounds,
+                    bounds: Rectangle {
+                        x: bounds.x + style.offset_left as f32,
+                        y: bounds.y,
+                        width: bounds.width - style.offset_right as f32,
+                        height: bounds.height,
+                    },
                     background: style
                         .background
                         .unwrap_or(Background::Color(Color::TRANSPARENT)),
--- a/crates/widgets/src/renderer/table_row.rs
+++ b/crates/widgets/src/renderer/table_row.rs
@@ -18,9 +18,10 @@ where
         style_sheet: &Self::Style,
         content: &Element<'_, Message, Self>,
         viewport: &Rectangle,
+        custom_bounds: &Rectangle,
     ) -> Self::Output {
         let bounds = layout.bounds();
-        let is_mouse_over = bounds.contains(cursor_position);
+        let is_mouse_over = custom_bounds.contains(cursor_position);
         let content_layout = layout.children().next().unwrap();
 
         let style = if is_mouse_over {
@@ -39,7 +40,7 @@ where
                         x: bounds.x + style.offset_left as f32,
                         y: bounds.y,
                         width: bounds.width - style.offset_right as f32,
-                        height: bounds.height,
+                        height: custom_bounds.height,
                     },
                     background: style
                         .background
--- a/crates/widgets/src/renderer/table_row.rs
+++ b/crates/widgets/src/renderer/table_row.rs
@@ -1,12 +1,12 @@
-use crate::widget::table_row;
-use crate::style::table_row::{StyleSheet};
-use iced_graphics::{Backend, Primitive, Renderer};
-use iced_native::{Background, Element, Layout, Point, Color, Rectangle};
+use crate::style::table_row::StyleSheet;
 
+use crate::widget::table_row;
+use iced_graphics::{Backend, Primitive, Renderer};
+use iced_native::{Background, Color, Element, Layout, Point, Rectangle};
 
 impl<B> table_row::Renderer for Renderer<B>
-    where
-        B: Backend
+where
+    B: Backend,
 {
     type Style = Box<dyn StyleSheet>;
 
@@ -19,7 +19,6 @@ impl<B> table_row::Renderer for Renderer<B>
         content: &Element<'_, Message, Self>,
         content_layout: Layout<'_>,
     ) -> Self::Output {
-
         let is_mouse_over = bounds.contains(cursor_position);
 
         let style = if is_mouse_over {
@@ -28,12 +27,8 @@ impl<B> table_row::Renderer for Renderer<B>
             style_sheet.style()
         };
 
-        let (content, mouse_interaction) = content.draw(
-            self,
-            &defaults,
-            content_layout,
-            cursor_position
-        );
+        let (content, mouse_interaction) =
+            content.draw(self, &defaults, content_layout, cursor_position);
 
         (
             if style.background.is_some() || style.border_width > 0 {
@@ -47,10 +42,9 @@ impl<B> table_row::Renderer for Renderer<B>
                     border_color: style.border_color,
                 };
 
-                    Primitive::Group {
-                        primitives: vec![background, content],
-                    }
-
+                Primitive::Group {
+                    primitives: vec![background, content],
+                }
             } else {
                 content
             },
--- a/crates/widgets/src/renderer/table_row.rs
+++ b/crates/widgets/src/renderer/table_row.rs
@@ -33,7 +33,7 @@ where
             content.draw(self, &defaults, content_layout, cursor_position, viewport);
 
         (
-            if is_mouse_over && (style.background.is_some() || style.border_width > 0.0) {
+            if style.background.is_some() {
                 let background = Primitive::Quad {
                     bounds: Rectangle {
                         x: bounds.x + style.offset_left as f32,
--- a/crates/widgets/src/renderer/table_row.rs
+++ b/crates/widgets/src/renderer/table_row.rs
@@ -13,13 +13,15 @@ where
     fn draw<Message>(
         &mut self,
         defaults: &Self::Defaults,
-        bounds: Rectangle,
+        layout: Layout<'_>,
         cursor_position: Point,
         style_sheet: &Self::Style,
         content: &Element<'_, Message, Self>,
-        content_layout: Layout<'_>,
+        viewport: &Rectangle,
     ) -> Self::Output {
+        let bounds = layout.bounds();
         let is_mouse_over = bounds.contains(cursor_position);
+        let content_layout = layout.children().next().unwrap();
 
         let style = if is_mouse_over {
             style_sheet.hovered()
@@ -28,10 +30,10 @@ where
         };
 
         let (content, mouse_interaction) =
-            content.draw(self, &defaults, content_layout, cursor_position);
+            content.draw(self, &defaults, content_layout, cursor_position, viewport);
 
         (
-            if style.background.is_some() || style.border_width > 0 {
+            if style.background.is_some() || style.border_width > 0.0 {
                 let background = Primitive::Quad {
                     bounds,
                     background: style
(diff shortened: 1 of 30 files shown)


tarkah:
```
> new to rust
> wanted easy feature
> adds custom widget
```

Wow! This is really awesome! Great stuff @kawashita86. You definitely didn't choose the easiest thing to implement :P

Can you run `rustfmt` and fix the lints with `clippy``? I'll dig into this more later today, but at first glance looks like the right approach.

tarkah:
Also, can you rebase to development branch? We upgraded our iced dependencies to 0.2 release and there are also some small api changes inside widget due to that.

kawashita86:
Sure, let me rebase and push it

casperstorm:
First of all, great first contribution. Wow.
I just tried it out and it will be a nice addition, very requested.

Heres a few things from a UI perspective:

- We have a left and right padding around content, I think we should wrap your highlightable row into a new row maybe so we don't highlight this padding.
- Colors is almost there, right now I think it is a little too distracting. I would like you to maybe try using the faded color for border as well, so its the same as your highlighted background. This is of cause just me suggesting.

kawashita86:
@casperstorm Thanks again for the kind words, glad that the addition is useful.
I completely agree with your feedback and I will try to get it done with some more coding, i will push them to this branch as soon as i've got them fixed.

crates/widgets/src/widget/table_row.rs
```diff
@@ -182,6 +183,8 @@ where
             renderer,
             clipboard,
         );
+
+        event::Status::Ignored
```
tarkah: You should return the status from `self.content.on_event` since it needs to bubble up

tarkah (changes requested):

crates/widgets/src/widget/table_row.rs
kawashita86: Thanks for the suggestion in the next commit i'm fixing it.

kawashita86:
Hello,
with the last commit i should have improved a bit the layout of the table_row widget:
- reduced the hovered contrast by changing the opacity of the border to a softer value
- added new property to the style of the table_row that define the offset of the row within the bounds of the layout, so now the hovered state should not get out of bounds and can be controlled.
- now the "on_event" function return the self.content.on_event status as suggested by @tarkah

casperstorm:
> with the last commit i should have improved a bit the layout of the table_row widget:

Awesome job once again. It looks nice. Here is a side-by-side from development and your branch:

highlight

- Would you try to work on getting us back to 1px between rows? I think it looks cleaner but it might not be easy so lets see what options we have to get there.
- I still feel we can optimize the coloring a bit. Optimal I would like us to keep the addon highlighting color similar to when you hover a column title: border color and background is the same. I feel it will create a better harmony.
- When you see a highlightable row/cell on our smart phone we are used to being able to click it. This is not the case right now. We have to click the title in order to expand it. Going into the future I would like us to consider if its possible that the whole row is clickable, while still being able to add buttons like "Update" which wouldn't trigger the cell click handler.

Again small points of feedback you can pick up if you have the time. Again thank you for doing this and welcome to the team :)

kawashita86:
Based on the suggestion i've decided to rework a bit the idea behind a current table row:
the are now two new style implementation  "HoverableBrightForegroundContainer" and "HoverableForegroundContainer" these are similar to the old "BrightForegroundContainer" and "NormalForegroundContainer" but they don't have a background set, so now the table_row is in charge of displaying the correct background color for the entire row, both in hovered and normal state. 
This should also make it easier to create further styling by having a single entrypoint, and should adhere more to the original layout design.

> When you see a highlightable row/cell on our smart phone we are used to being able to click it. This is not the case right now. We have to click the title in order to expand it. Going into the future I would like us to consider if its possible that the whole row is clickable, while still being able to add buttons like "Update" which wouldn't trigger the cell click handler.

I totally agree with your idea, I will study a bit this and see if there is a clean approach to solve the event propagation and capture as I fear that it can get easily messy without a good implementation.

tarkah:
@kawashita86 you can check the return of `self.content.on_event` and if it is `Status::Ignored`, then you check query the event to see if it's a left click. If so, you can append an `on_click` message into `messages`. You should add on `on_click` builder method similar to the `on_resize` one I have for the header widget. I hope that makes sense.

kawashita86:
@tarkah Thanks for pointing out a way to achieve the wanted behaviour.
I was working on something really similar, but i was slowed down a bit since when the row is expanded (on MyAddon tab) the pickup list for the release channel is propagating back the `event::Status::Ignored` creating a weird behaviour of the row opening and closing any time you interact with the picklist.
I've decided to use  a `custom_bounds` Rectangle inside the `on_event` that is limited to the bounds of the closed row instead of the expanded one.
I will wait for your feedback to move on.

crates/widgets/src/widget/table_row.rs
```diff
@@ -72,6 +76,12 @@ where
         self
     }
 
+    /// Sets the maximum height of the [`TableRow`] in pixels.
```
tarkah, line 79: I'd reword this to explain this is the max height of the "clickable" area

tarkah (changes requested):
This looks really good! I'd change the the hovered style to only trigger based on the custom bounds as well. That way the expanded row doesn't show as highlighted when interacting with it. Then the pick-list issue isn't even a problem. If we want the expanded row to have different coloring, we should handle that styling based on the fact it's expanded, not based on the "hover" of this row.

tarkah:
I'm running into a couple weird bugs w/ this PR. The first, there is a visual artifact at the top between the flavor selector and version:

image

The second.. if I left click on the scrollbar, it gets "stuck" and follows my mouse movement around. What I expect is happening is the scrollbar captures the left click, but doesn't capture the release, because our row is taking precedence maybe? Similar to how you have the event capturing only taking place after a certain height, can you lock the event capturing to be only inside the bounds of the actual rectangle?

crates/widgets/src/renderer/table_row.rs
```diff
@@ -0,0 +1,61 @@
...
+        };
+
+        let (content, mouse_interaction) =
+            content.draw(self, &defaults, content_layout, cursor_position, viewport);
+
+        (
+            if style.background.is_some() {
+                let background = Primitive::Quad {
+                    bounds: Rectangle {
+                        x: bounds.x + style.offset_left as f32,
```
tarkah, line 39: We shouldn't hardcode this offset to the implementation of this widget. What if we want to use this widget, but have it highlight the entire thing? We should just wrap this widget in a Row widget and apply the padding there by pushing some Space on each side of it.

tarkah (changes requested):

crates/widgets/src/renderer/table_row.rs
tarkah: Ahh, I see we can specify it with the style, so I guess not that big of a deal. I'm ok leaving it this way.

tarkah: And when I think about it more, we would need this anyway to work well with the nested scrollable since we don't want the scrollbar to get highlighted when hovering over it, so a offset for the styling would be needed in that case

tarkah: Omg I'm retarded, disregard everything I said. This is inside the scollable, not the other way around :P

tarkah:
Ok, here is a fresh comment since I was all over the place :P So the Scrollable widget is bugged, in that it resolves nested events _before_ the events on the scrollbar, even though the scrollbar should take precedence over any widget nested inside the scrollable since it's technically the top layer of the nested contents. I'll open a PR w/ Iced so they can fix this behavior.

The fix for us is we need to not return `Status::Captured` on left mouse button release of our TableRow widget as it prevents the scrollbar from resolving on that event, which thing locks the scrollbar as captured.

tarkah:
Opened issue here: https://github.com/hecrj/iced/issues/649

kawashita86:
@tarkah first of all thank you for taking time to go through everything in details, as you guessed I've already implemented a public method to set the height of the row used for the custom bound, probably `inner_row_height` is not the most fortunate choice ever.
I'm not happy with the idea of having to set manually the height of a property that is used mostly for internal behaviour, but it was the best idea i could find to solve the pick-list issue without creating a monster.
So as you suggest i will change the current render method to apply the hovered style to the custom_bounds only and change the event::Status::Captured to circumvent the current iced bug.

tarkah:
Great! An for the `menu_container` can you please revert the style for those back to `NormalForegroundContainer`? I believe that's what's causing the visual bug in the top menu area. I assume this was overridden with a find / replace?

kawashita86:
Ok so i've committed the latest changes, i should also have reverted the old `menu_container` style to the previous `BrightForegroundContainer`.

src/gui/element.rs
```diff
@@ -1715,7 +1723,7 @@ pub fn menu_container<'a>(
     let version_container = Container::new(version_text)
         .center_y()
         .padding(5)
-        .style(style::NormalForegroundContainer(color_palette));
+        .style(style::HoverableForegroundContainer(color_palette));
```
tarkah, line 1726: Still one that needs to be changed here. Is everything else good to go outside this change? If so, I'll do a final test before merging.

tarkah (changes requested):

kawashita86 (commented):
I've reviewed all the request opened

tarkah (approved):
Great changes. This looks really nice! @casperstorm all looks good to me

casperstorm (changes requested):
One last thing I have is to convert addon title from a Button to a Text. We don't need to have that now that we can (FINALLY!) press the row it self to expand. Should be super easy, basically just convert this to a `Text::new` (or see how its done for `installed_version`). The relevant code is right here:

https://github.com/casperstorm/ajour/blob/90feb56bc31b0d6d1fa978513f41957d0e3a402e/src/gui/element.rs#L739-L740

@kawashita86 you have done amazingly. I can't believe how great this PR has turned out to be. My testing found no issues.

kawashita86:
> One last thing I have is to convert addon title from a Button to a Text. We don't need to have that now that we can (FINALLY!) press the row it self to expand. Should be super easy, basically just convert this to a `Text::new` (or see how its done for `installed_version`). The relevant code is right here:
> 
> https://github.com/casperstorm/ajour/blob/90feb56bc31b0d6d1fa978513f41957d0e3a402e/src/gui/element.rs#L739-L740
> 

Ok, I've reworked a bit how a row is built as you suggested:
- Removed the button from both the "My Addons" tab and the "Catalog", now they are simple `Text` elements.
- Removed the ` website_state: button::State` from `CatalogRow` as it was not used anymore.
- Added in the `table_row` renderer the` mouse::Interaction::Pointer` when the mouse is in hovered state, to make it easier to understand that there's a possible action by clicking the row.

It should be everything now.

casperstorm (approved):

casperstorm:
@kawashita86 Amazing, just tested and everything works nicely.
Do you want to update CHANGELOG.md or should I?

kawashita86:
@casperstorm Glad that everything works fine!! Please go on with the Changelog entry, i'm SO bad at writing commit messages and changelog  that i'm always happy when someone else takes the task.

PR #404, opened 2020-12-01
chore: update tukui endpoint

casperstorm:
## Proposed Changes
  - Switch to Hub api, like on Curse.

## Checklist

- [x] Switch to production
- [x] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/crates/core/src/config/wow.rs
+++ b/crates/core/src/config/wow.rs
@@ -49,14 +49,6 @@ impl Flavor {
         }
     }
 
-    /// Returns flavor `String` in Tukui format
-    pub(crate) fn tukui_format(self) -> String {
-        match self {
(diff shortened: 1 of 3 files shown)


crates/core/src/config/wow.rs
```diff
@@ -49,6 +49,14 @@ impl Flavor {
         }
     }
 
+    /// Returns flavor `String` in Tukui format
+    pub(crate) fn tukui_format(self) -> String {
```
tarkah, line 53: Are we using this anywhere other than building the API URL? If not, we might want to move this to a helper in that tukui backend module and change `classic` to `classic-addon` and reference this for the classic match as well

casperstorm: Yeah good point, also changed for classic addons.

tarkah (approved):
Looks good! I think we can merge this once we're sure the wowhub api is stable.

PR #405, opened 2020-12-01
feat: simplify changelogs

casperstorm:
## Proposed Changes
  - Not happy with how complex the changelog system is. Also not happy with poorly we are parsing it. It will never be perfect to what I would like because of the different formats. This change will remove the parsed changelogs and add a "Changelog" button in details which opens directly to the relevant page with a changelog.

## Checklist

- [x] Cleanup
- [ ] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux


tarkah (approved):
Looks good! Removes a lot of unneeded complexity

PR #409, opened 2020-12-03
add self update release channel

tarkah:
## Proposed Changes
  - Add release channel for self update (Release vs PreRelease)
  - PreRelease will show updates in Ajour for normal releases, as well as prerelease releases
  - Add picklist in settings to switch between channels

## Checklist

- [x] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/crates/core/src/utility.rs
+++ b/crates/core/src/utility.rs
@@ -1,3 +1,4 @@
+use crate::config::SelfUpdateChannel;
 use crate::error::DownloadError;
 #[cfg(target_os = "macos")]
 use crate::error::FilesystemError;
@@ -28,6 +29,7 @@ pub(crate) fn strip_non_digits(string: &str) -> Option<String> {
 #[derive(Debug, Deserialize, Clone)]
 pub struct Release {
     pub tag_name: String,
+    pub prerelease: bool,
     pub assets: Vec<ReleaseAsset>,
     pub body: String,
 }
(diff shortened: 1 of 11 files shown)


tarkah:
We need to fix our self update logic to not break when triggered while Ajour is still running from an instance that was self updated. Since we will want to force a recheck of releases when channel is changed and a self update can now technically happen after one already did.

tarkah:
@casperstorm I noticed when testing this that the "About" tab dissapears when an update is available.

casperstorm:
> @casperstorm I noticed when testing this that the "About" tab dissapears when an update is available.

That's actually on purpose!
If it's unintuitive then we can look at changing it.

tarkah:
> 
> 
> > @casperstorm I noticed when testing this that the "About" tab dissapears when an update is available.
> 
> That's actually on purpose!
> If it's unintuitive then we can look at changing it.

No I think it makes sense, I just never realized before so wanted to make sure it was intentional

crates/core/src/utility.rs
```diff
@@ -87,6 +87,12 @@ pub async fn download_update_to_temp_file(
         .unwrap()
         .to_str()
         .unwrap()
+        // Trim `temp_` from the `current_exe` path in case this self update process
```
casperstorm, line 90: Nice documented.

tarkah: It's hard to make it make sense without way more context, but better than nothing :)

PR #411, opened 2020-12-04
Chore/ci improvements

tarkah:
## Proposed Changes
  - Only run CI when program files change. It doesn't need to run when we update README, CHANGELOG, etc.
  - Add caching to improve build times and bust cache if Cargo.lock has changed

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux


casperstorm (approved):

casperstorm:
LGTM!

PR #417, opened 2020-12-07
Fix for bug on #416

### LINKED ISSUE #416: [RC] Selecting an item from a pick_list which expands over a table_row.rs will fire `on_press` event
**Describe the bug**
When you select an item from a `pick_list` inside the Catalog it will fire `on_press` on the cell below the `pick_list`.
This is only happning after https://github.com/casperstorm/ajour/commit/b5ab26f4875b19fb725e7324f98fd7f6f504aefa. Not in production yet.

**To Reproduce**
1. Go to Catalog
2. Press eg. All Catagories
3. Select a category
4. Notice the addon underneath that catagory you pressed will open up in your browser.

What happens is that the press apparently goes through the `pick_list` and the `table_row` gets it, and then calls `on_press` for the addon. The relevant code is here.

https://github.com/casperstorm/ajour/blob/2a4d2f3d069fe2781fdd0953614b2bed97b5ffda/crates/widgets/src/widget/table_row.rs#L220-L222

@kawashita86 or @tarkah – is this something you can take a stab at?

kawashita86:
Resolves #416 
Changed `Event::Mouse(mouse::Event::ButtonReleased(mouse::Button::Left)) `on `table_row`, to `Event::Mouse(mouse::Event::ButtonPressed(mouse::Button::Left))` to avoid having uncaptured event from other widget

### DIFF
--- a/crates/widgets/src/widget/table_row.rs
+++ b/crates/widgets/src/widget/table_row.rs
@@ -224,10 +224,7 @@ where
                 }
                 status_from_content
             }
-            _ => {
-                println!("{:?}", event);
-                status_from_content
-            }
+            _ => status_from_content,
         }
     }
 
--- a/crates/widgets/src/widget/table_row.rs
+++ b/crates/widgets/src/widget/table_row.rs
@@ -207,7 +207,7 @@ where
         );
         match status_from_content {
             event::Status::Ignored => {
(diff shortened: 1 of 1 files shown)


casperstorm:
Thanks @kawashita86.
Btw are you on Discord? We have a server where we are talking: https://discord.com/invite/ajour

PR #419, opened 2020-12-08
add fuzzy match to catalog search

### LINKED ISSUE #398: Improve search matching
Searching the catalog for "handynotes shadowlands" return zero results. Searching for "handynotes: shadowlands" returns two results.

Search should ignore punctuation when determining matches.

**Additional context**
image
image

tarkah:
Resolves #398

## Proposed Changes
  - When user searches for an addon in the catalog, use a fuzzy match instead of basic string contains logic.
  - Results are sorted by fuzzy match score by default. Columns can still be sorted after the fact, but typing a new character into the query box resets the sort back to fuzzy match score.

## Checklist

- [x] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ and `Removed`.
 
 ## [Unreleased]
 
+### Changed
+
+- Catalog search now uses a fuzzy match for better searching of the catalog.
+
 ## [0.5.4] - 2020-12-07
 
 ### Added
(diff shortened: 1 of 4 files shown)


tarkah:
Searching is super fast in `--release` mode, but do note it feels super sluggish in dev

casperstorm (approved):
Amazing addition @tarkah.
Works great for me on both macOS and Windows.

LGTM.

PR #420, opened 2020-12-09
Fix/column migration issue

### LINKED ISSUE #361: Catalog view, column "addon" does not resize
**Describe the bug**
When browsing the catalog, one can't resize the column in which the name of the addon appear. It, instead, resizes the "description" column

**To Reproduce**
1. Catalog
2. Drag the right edge of column to resize
3. bug

**Expected behavior**
Resizes the correct column

**Screenshots**
Nop

**Software involved**
Tested on Win10, Ajour 0.5.1

tarkah:
Resolves #361

## Proposed Changes
  - Always for the "Title" column to be Fill width, regardless of what's set in the config.

## Checklist

- [x] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ and `Removed`.
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix a bug where users upgrading from an older version of Ajour might have incorrect
+  behavior when trying to resize a column in the Catalog
+
 ## [0.5.4] - 2020-12-07
 
 ### Added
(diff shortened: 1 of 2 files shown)


tarkah:
Tested this by setting Title column to a set width in the config on development and experienced the resizing issues. Switched to this branch and everything worked perfectly, and the config gets saved back with the appropriate `~`

tarkah:
I've also tested what would happen if two different columns happen to be set to `~`. The good news is we already had logic in place to only ever allow the Title columns to stay as Fill when a resize occurs, so the other one will have it's width computed on next resize and get saved correctly to the config.

casperstorm (approved):
LGTM. This bug annoyed me.

Issue #231, opened 2020-10-11
Set default preferred release type


KestrelWild:
Would love to have the ability to set my preferred release type for all addons as beta, or at least, for all newly installed addons.

casperstorm:
Hi @KestrelWild,

I talked about this feature a [little while ago](https://github.com/casperstorm/ajour/issues/185) so I am just gonna quote myself:

> We talked about adding this, but I decided it was too niche.
For now you'll have to do it per addon, but that will however be saved so you only have to do it once.
>
> I'll might reconsider it once we are more feature complete, but for now this will not be added to the roadmap.

We are a little further now, so I did note it down in my personal notes 🙂

Issue #185, opened 2020-10-02
Mass release channel switch


Darhel:
Its possible to add mass release channel switch button, to switch all addons at one click? 
Thanks anyway for great software.

casperstorm:
Thank you so much for contributing with a feature request.

We talked about adding this, but I decided it was too niche.
For now you'll have to do it per addon, but that will however be saved so you only have to do it once.

I'll might reconsider it once we are more feature complete, but for now this will not be added to the roadmap.

PR #428, opened 2020-12-10
feat: global addon release channel

### LINKED ISSUE #236: Global setting to automatically select actual latest version
Manually going through every addon to check whether there's an alpha/beta available and determine if it's newer than the release version is incredibly tedious.

Drop-down to select globally preferred release type, and recognize newer release/beta versions as upgrades to an alpha version

CurseBreaker does this.

casperstorm:
Resolves #236 #185 #231

## Proposed Changes
  - We will create a way to control release channels at a global level.

## Checklist

- [x] Update all addons which is `default` on selection.
- [x] Ensure we download the correct release from catalog. With fallbacks.
- [x] Migration (All `stable` goes to `default`).
- [x] Tested on Windows
- [x] Tested on MacOS
- [x] Tested on Linux

### DIFF
--- a/crates/core/src/repository/mod.rs
+++ b/crates/core/src/repository/mod.rs
@@ -213,13 +213,15 @@ impl Ord for RemotePackage {
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Deserialize, Serialize, Hash, PartialOrd, Ord)]
 pub enum ReleaseChannel {
+    Default,
     Stable,
     Beta,
     Alpha,
 }
 
 impl ReleaseChannel {
-    pub const ALL: [ReleaseChannel; 3] = [
+    pub const ALL: [ReleaseChannel; 4] = [
+        ReleaseChannel::Default,
         ReleaseChannel::Stable,
         ReleaseChannel::Beta,
         ReleaseChannel::Alpha,
@@ -228,7 +230,7 @@ impl ReleaseChannel {
 
 impl Default for ReleaseChannel {
     fn default() -> ReleaseChannel {
-        ReleaseChannel::Stable
+        ReleaseChannel::Default
     }
 }
 
@@ -238,6 +240,7 @@ impl std::fmt::Display for ReleaseChannel {
             f,
             "{}",
             match self {
+                ReleaseChannel::Default => "Default",
                 ReleaseChannel::Stable => "Stable",
                 ReleaseChannel::Beta => "Beta",
                 ReleaseChannel::Alpha => "Alpha",
--- a/crates/core/src/repository/mod.rs
+++ b/crates/core/src/repository/mod.rs
@@ -227,6 +227,14 @@ impl GlobalReleaseChannel {
         GlobalReleaseChannel::Beta,
         GlobalReleaseChannel::Alpha,
     ];
+
+    pub fn convert_to_release_channel(&self) -> ReleaseChannel {
+        match self {
+            GlobalReleaseChannel::Stable => ReleaseChannel::Stable,
+            GlobalReleaseChannel::Beta => ReleaseChannel::Beta,
+            GlobalReleaseChannel::Alpha => ReleaseChannel::Alpha,
+        }
+    }
 }
 
 impl Default for GlobalReleaseChannel {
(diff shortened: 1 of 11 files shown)


crates/core/src/repository/mod.rs
```diff
@@ -228,7 +271,7 @@ impl ReleaseChannel {
 
 impl Default for ReleaseChannel {
     fn default() -> ReleaseChannel {
-        ReleaseChannel::Stable
+        ReleaseChannel::Default(GlobalReleaseChannel::Stable)
```
casperstorm, line 274: @tarkah I am running in circles here.
If we don't embed `GlobalReleaseChannel` here we have to pass a reference to `Config` every time, and it's used A LOT of places.
If we pass it like i have now we would need to set it on every parse. I would like some help/feedback here. What is the right play?

src/gui/update.rs
```diff
@@ -1488,6 +1488,13 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
                 Message::LatestRelease,
             ));
         }
+        Message::Interaction(Interaction::PickGlobalReleaseChannel(channel)) => {
+            log::debug!("Interaction::PickGlobalReleaseChannel({:?})", channel);
+
+            ajour.config.addons.global_release_channel = channel;
+
+            let _ = ajour.config.save();
+        }
```
casperstorm, line 1497: ... And if we do `ReleaseChannel::Default(GlobalReleaseChannel)`  what should happen here? Loop and set?

casperstorm:
@tarkah take a look, im still working on it but now i am passing `global_release_channel` as we talked about. How do we like this?

tarkah:
@casperstorm looks good! A couple things I noticed from testing...

This line pretty much forces curse addons "Stable" or whatever version is installed... I think we should remove this? 

https://github.com/casperstorm/ajour/blob/424a14c01ef5e32f5e42f2f733c37a02e8a5954b/src/gui/update.rs#L417-L424

Also, when changing the global to Alpha, for example, addons that now have an alpha update don't show as "updatable" until we do a "Refresh". We should make sure this shows up once the global selector is changed.

casperstorm:
> @casperstorm looks good! A couple things I noticed from testing...
> This line pretty much forces curse addons "Stable" or whatever version is installed... I think we should remove this?
> https://github.com/casperstorm/ajour/blob/424a14c01ef5e32f5e42f2f733c37a02e8a5954b/src/gui/update.rs#L417-L424

This should be done: https://github.com/casperstorm/ajour/pull/428/commits/1288f5896c9015bbebb333fdc53dbaf80efae3d0

casperstorm:
> Also, when changing the global to Alpha, for example, addons that now have an alpha update don't show as "updatable" until we do a "Refresh". We should make sure this shows up once the global selector is changed.

Fixed: https://github.com/casperstorm/ajour/pull/428/commits/389a7044d90e81e183f59a3ff5534e7c43b9f31d

casperstorm:
Can you take a look again, @tarkah?
Also, what do we do with: 
https://github.com/casperstorm/ajour/blob/389a7044d90e81e183f59a3ff5534e7c43b9f31d/crates/core/src/addon.rs#L517-L527

Is this safe to delete? I think it is, havn't found it affect the sorting.

src/gui/update.rs
```diff
@@ -414,13 +414,8 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
                             {
                                 a.release_channel = *release_channel;
                             } else {
-                                // Else we try to determine the release_channel based of installed version.
```
tarkah, line 409: I think we just delete the else block here since `ReleaseChannel::Default` is the `Default` and what gets set when we create an addon.

tarkah:
> Can you take a look again, @tarkah?
> Also, what do we do with:
> 
> https://github.com/casperstorm/ajour/blob/389a7044d90e81e183f59a3ff5534e7c43b9f31d/crates/core/src/addon.rs#L517-L527
> 
> Is this safe to delete? I think it is, havn't found it affect the sorting.

I believe it is safe to delete.

tarkah:
@casperstorm changes work great! I noticed if you change something from `Default` to one of the channels, then back to `Default`, it gets saved in the config as `Default` for the release channel. I think we should just remove the entry instead of saving it as Default in the config file.

Other than that, everything tested as working great!

casperstorm:
This should be good to go now.

PR #429, opened 2020-12-11
remove addon cache entries on missing addon folders

tarkah:
## Proposed Changes
  - Remove addon cache entries when it's folders no longer exist on the filesystem

## Checklist

- [x] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ and `Removed`.
 - Fix a bug where users upgrading from an older version of Ajour might have incorrect
   behavior when trying to resize a column in the Catalog
 - Ignored addons are now sorted correctly again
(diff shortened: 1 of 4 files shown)


tarkah:
@casperstorm I've added logic to remove any addon cache entries during parse if it's addon folders are not completely on the filesystem anymore.

I've also added logic to remove all fingerprint entries for a flavor if that flavors AddOn folder is missing or empty. We already remove single entries for missing folders, but since we return early from the parse function on empty or missing Addon folder, I covered that base here as well.

casperstorm (approved):
LGTM.
Great comments, and easy to follow. Tested on macOS.

PR #436, opened 2020-12-17
chore: fix for selecting default release channel

casperstorm:
Resolves issue where if you go from `stable` to `default` it wouldn't remove `stable` from config.


tarkah (approved):
looks good!

PR #437, opened 2020-12-18
add padding back to titles

tarkah:
There was a regression in 0.5.4 once highlight rows was enabled that removed padding from the Title of My Addons and Catalog. The text for these used to be wrapped in a Button, which by default has a padding of 5 on it's contents. I fixed this by adding the padding of 5 back on the text's container. This also fixes the regression where the title would visibly "wrap" without this padding.

You can see the default padding of a button here:

https://github.com/hecrj/iced/blob/master/graphics/src/widget/button.rs#L24

### Before

Screenshot from 2020-12-17 15-52-26

### After

Screenshot from 2020-12-17 16-04-22

## Checklist

- [x] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/src/gui/element/my_addons.rs
+++ b/src/gui/element/my_addons.rs
@@ -147,8 +147,6 @@ pub fn data_row_container<'a, 'b>(
     {
         let title = Text::new(addon.title()).size(DEFAULT_FONT_SIZE);
 
-        if release_package.is_some() {}
-
         let mut title_row = Row::new().push(title).spacing(5).align_items(Align::Center);
 
         if addon.release_channel != ReleaseChannel::Default {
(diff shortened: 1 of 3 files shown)


src/gui/element/my_addons.rs
```diff
@@ -147,8 +147,6 @@ pub fn data_row_container<'a, 'b>(
     {
         let title = Text::new(addon.title()).size(DEFAULT_FONT_SIZE);
 
-        if release_package.is_some() {}
```
tarkah, line 150: Lol I noticed this useless statement here when updating :P

casperstorm:
LGTM

PR #440, opened 2020-12-20
fix: linux resizable

### LINKED ISSUE #427: Application no resizable on wayland since 0.5.4
**Describe the bug**
Application window no resizable on wayland since 0.5.4

**To Reproduce**
Launch app on wayland, try to resize, you cannot

It worked on 0.5.3

**Expected behavior**
We should ensure resizing works

**Screenshots**
Capture d’écran du 2020-12-10 16-04-15

Also the maximize doesn't work neither

**Software involved**
 - OS: Archlinux
 - Ajour version: 0.5.4

**Additional context**

**Log Output**
I take a look but there is no error or warning, only valid info

casperstorm:
Resolves #427 

## Proposed Changes
  - Disabled minimum window size on Linux because it causes the window to become non-resizable.

## Checklist

- [ ] Tested on Windows
- [X] Tested on MacOS
- [x] Tested on Linux

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@ and `Removed`.
 
 ### Fixed
 
+- Removed minimum window size on Linux. This fixed a issue where the application would
+  not be resizable.
 - Certain CF addons weren't fingerprinted correctly due to a bug in the fingerprinting
   logic for splitting on lines.
 - Fix a bug where users upgrading from an older version of Ajour might have incorrect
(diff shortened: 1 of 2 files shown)


casperstorm:
@tarkah @nerzhul Can you guys this this branch to see if it fixes the issue?

tarkah:
I'm not using Wayland so I can't test, but the change looks correct.

tarkah:
@lub can you confirm this branches resolves the resize issue?

casperstorm:
Confirmed to work: https://github.com/casperstorm/ajour/issues/444#issuecomment-749377010

PR #448, opened 2020-12-21
include Interface root in AddOns backup path 

### LINKED ISSUE #433: Backup Addons at Interface level instead of Addons folder only
Some addon packs and I use the Interface folder for custom textures rather than the AddOns one. 

I would really like is for the backup feature to not only backup the addons folder, but the parent one instead.

**Additional context**

tarkah:
Resolves #433

## Proposed Changes
- `Interface` folder root is included in the zip backup when `AddOns` is selected.
  Some users store proprietary data alongside the `AddOns` folder that they'd like
  included during backup.

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [x] Tested on Linux

### DIFF
--- a/src/gui/update.rs
+++ b/src/gui/update.rs
@@ -1085,7 +1085,7 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
                 if ajour.config.backup_addons {
                     let addon_dir = ajour.config.get_addon_directory_for_flavor(flavor).unwrap();
 
-                    // Backup starting with `Interface` fodler as some users save
+                    // Backup starting with `Interface` folder as some users save
                     // custom data here that they would like retained
                     if let Some(interface_dir) = addon_dir.parent() {
                         if interface_dir.exists() {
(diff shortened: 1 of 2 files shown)


src/gui/update.rs
```diff
@@ -1085,8 +1085,12 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
                 if ajour.config.backup_addons {
                     let addon_dir = ajour.config.get_addon_directory_for_flavor(flavor).unwrap();
 
-                    if addon_dir.exists() {
-                        src_folders.push(BackupFolder::new(&addon_dir, wow_dir));
+                    // Backup starting with `Interface` fodler as some users save
```
casperstorm, line 1088: Small spelling issue "fodler" -> "folder"

casperstorm (changes requested):
Small spelling issue, otherwise LGTM!

src/gui/update.rs
tarkah: good catch!

PR #449, opened 2020-12-21
add alternating row colors 

tarkah:
## Proposed Changes
  - Row colors now alternate by default for better accessibility. This can be turned
  off in settings under the UI section's "Alternate Row Colors" checkbox.

## Checklist

- [x] Tested on Windows
- [x] Tested on MacOS
- [x] Tested on Linux

### DIFF
--- a/src/gui/mod.rs
+++ b/src/gui/mod.rs
@@ -116,6 +116,7 @@ pub enum Interaction {
     ToggleBackupFolder(bool, BackupFolderKind),
     PickSelfUpdateChannel(SelfUpdateChannel),
     PickGlobalReleaseChannel(GlobalReleaseChannel),
+    AlternatingRowColorToggled(bool),
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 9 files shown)


src/gui/mod.rs
```diff
@@ -446,6 +447,12 @@ impl Application for Ajour {
                         ExpandType::None => false,
                     };
 
+                    let is_odd = if self.config.alternating_row_colors {
+                        Some(idx % 2 != 0)
+                    } else {
+                        None
+                    };
```
casperstorm, line 454: Clever :)

casperstorm (commented):
LGTM. This I really like a lot.
I'll quickly go through all themes and adjust anything if needed.

src/gui/mod.rs
tarkah: :D

casperstorm (approved):

tarkah:
> LGTM. This I really like a lot.
> I'll quickly go through all themes and adjust anything if needed.
> 
> Edit: I actually think they all look good. Not sure I want to do any updates on them. Ayu for example does blend quite a lot with bg, but still is much more readable. What do you think? I actually like it as it is.

I think it's fine. Users needing more contrast can always create a custom theme.

PR #450, opened 2020-12-21
Change/remote version changelog

tarkah:
## Proposed Changes
  - Make remote version in row clickable to open changelog website
  - Curse link has also been updated to open to the specific file instead of all files window

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [x] Tested on Linux


casperstorm (approved):
LGTM, I think some will be happy for this. :)

PR #452, opened 2020-12-22
feat: reset columns option

casperstorm:
## Proposed Changes
  - Adds a reset columns button in settings. 

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/src/gui/mod.rs
+++ b/src/gui/mod.rs
@@ -117,6 +117,7 @@ pub enum Interaction {
     PickSelfUpdateChannel(SelfUpdateChannel),
     PickGlobalReleaseChannel(GlobalReleaseChannel),
     AlternatingRowColorToggled(bool),
+    ResetColumns,
 }
 
 #[derive(Debug)]
@@ -213,6 +214,7 @@ pub struct Ajour {
     weak_auras_is_installed: bool,
     weak_auras_state: HashMap<Flavor, WeakAurasState>,
     aura_header_state: AuraHeaderState,
+    reset_columns_btn_state: button::State,
 }
 
 impl Default for Ajour {
@@ -271,6 +273,7 @@ impl Default for Ajour {
             weak_auras_is_installed: Default::default(),
             weak_auras_state: Default::default(),
             aura_header_state: Default::default(),
+            reset_columns_btn_state: Default::default(),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/gui/update.rs
```diff
@@ -230,6 +230,13 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
...
             return Ok(Command::perform(select_directory(), message));
         }
+        Message::Interaction(Interaction::ResetColumns) => {
+            log::debug!("Interaction::ResetColumns");
+
+            ajour.config.column_config = Default::default();
+            apply_config(ajour, ajour.config.clone());
+            let _ = &ajour.config.save();
+        }
         Message::Interaction(Interaction::OpenLink(link)) => {
```
casperstorm, line 240: @tarkah, currently when pressing the button and executing this code and closing and reopening Ajour the columns is reset, however I am wondering why it does not happen instantly and appear changed in GUI?

src/gui/mod.rs
```diff
@@ -1987,7 +1991,7 @@ async fn load_caches() -> Result<(FingerprintCache, AddonCache)> {
     Ok((fingerprint_cache, addon_cache))
 }
 
-fn apply_config(ajour: &mut Ajour, config: Config) {
+pub fn apply_config(ajour: &mut Ajour, config: Config) {
```
casperstorm, line 1994: How do you like this, @tarkah? I would need to access `apply_config` from `update`.

tarkah: Yeah I think that makes sense! No reason to reinvent the wheel

src/gui/update.rs
tarkah: Let me test and see if I can figure it out

tarkah:
@casperstorm see my commit, I've fixed it. The problem is when you set the config.column_settings to Default, it just creates an empty HashMap. So when you did the `apply_config` function, it didn't reset the internal states correctly.

The proper order is to reset all the internal column state structs to Default then serialize those Defaults back to the config file.

casperstorm:
> @casperstorm see my commit, I've fixed it. The problem is when you set the config.column_settings to Default, it just creates an empty HashMap. So when you did the `apply_config` function, it didn't reset the internal states correctly.
> 
> 
> 
> The proper order is to reset all the internal column state structs to Default then serialize those Defaults back to the config file.

Ah of cause. Let's get this into today's RC! Awesome you nailed it.

tarkah:
Sweet! This is good to merge then when you're ready

PR #454, opened 2020-12-22
WIP: anyhow -> eyre

meltinglava:
This change is the start to try and print colorized error messages. Usefull for stuff if we want to show errors like wrong paths to wow folder, of what we expect. This first commit will have no changes to the program.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -11,13 +11,13 @@ use ajour_core::fs::CONFIG_DIR;
 use ajour_core::utility::{remove_file, rename};
 
 #[cfg(target_os = "linux")]
-use anyhow::Context;
+use eyre::WrapErr;
 use std::env;
 use std::path::PathBuf;
 
 pub const VERSION: &str = env!("CARGO_PKG_VERSION");
 
-pub type Result<T, E = anyhow::Error> = std::result::Result<T, E>;
+pub type Result<T, E = eyre::Report> = std::result::Result<T, E>;
 
 pub fn main() {
     let opts_result = cli::get_opts();
@@ -89,7 +89,7 @@ pub fn main() {
 }
 
 /// Log any errors
-pub fn log_error(error: &anyhow::Error) {
+pub fn log_error(error: &eyre::Report) {
     log::error!("{}", error);
 
     let mut causes = error.chain();
--- a/src/main.rs
+++ b/src/main.rs
@@ -9,33 +9,25 @@ mod gui;
 
 use ajour_core::fs::CONFIG_DIR;
 use ajour_core::utility::{remove_file, rename};
+use color_eyre::eyre::{Report, Result, WrapErr};
 
 #[cfg(target_os = "linux")]
-use eyre::WrapErr;
 use std::env;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 pub const VERSION: &str = env!("CARGO_PKG_VERSION");
 
-pub type Result<T, E = eyre::Report> = std::result::Result<T, E>;
-
-pub fn main() {
+pub fn main() -> color_eyre::eyre::Result<()> {
+    color_eyre::install()?;
     let opts_result = cli::get_opts();
 
-    #[cfg(debug_assertions)]
-    let is_debug = true;
-    #[cfg(not(debug_assertions))]
-    let is_debug = false;
+    let is_debug = cfg!(debug_assertions);
 
     // If this is a clap error, we map to None since we are going to exit and display
     // an error message anyway and this value won't matter. If it's not an error,
     // the underlying `command` will drive this variable. If a `command` is passed
     // on the command line, Ajour functions as a CLI instead of launching the GUI.
-    let is_cli = opts_result
-        .as_ref()
-        .map(|o| &o.command)
-        .unwrap_or(&None)
-        .is_some();
+    let is_cli = opts_result.as_ref().map(|o| &o.command).is_ok();
 
     // This function validates whether or not we need to exit and print any message
     // due to arguments passed on the command line. If not, it will return a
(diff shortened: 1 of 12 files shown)


tarkah:
@meltinglava Thanks for the PR! I'm curious if you have any examples of what this change would look like for errors on the CLI commands?

meltinglava:
I wanted to test [color-eyre](https://github.com/yaahc/color-eyre). Have not had time to start looking at that yet. poped this draft here just to remind me to continue the work. Also happy to let others try as well.

tarkah:
Sounds good! Just ping me when you've got some more implemented and we can take a look :)

meltinglava:
I have now done all of the things to get eyre print all error messages in colors. There are a few things that needs to be looked at. Would like to have some sort of end report that say what happened (like what updated from what version to other version). Would also like to have more errors return `Result::Err`, rather than panics. 

One thing i do not know yet is logging. Espesially in `CLI` mode as i think its often not needed. We want to just throw out a report when we fail somehow. For now i just have it when running `CLI` its reporting the same as `release GUI`

casperstorm:
Thanks for the PR. However, I am not sure I understand what the benefits of this is? Could you explain that to me? 🙂

meltinglava:
This PR want to improve the output from the CLI part of ajour. 

I think what keeps me away from using ajour as a cli application are (besides the functionality differences) are that the logs given to you are verbose. All I want it to do is say addons x y and z was updated from version x to y. I think logs are to be used for debugging, not for user interaction (even thow the all the user can do is reading it).

The answare to what benefits this PR gives us depends on what the goal is for ajour, and espesially the cli. Are the gui and the cli eventually going to be equivalent in functionality? If this is a goal then we would like that bought of them has good "interfaces" for what they are. I think the GUI part of ajour are good. The cli on the other hand looks to be the first test for functionality that was just put there before there was any GUI at all. Would like this to be as good as the GUI, but I find that is not the case at this moment. 

Think what i would like in future would be to `ajour install weakauras3` and it just does it(maybe with some interactive elements if needed).

meltinglava:
@tarkah Here is an error for what an invalid url would give on install. What makes this not look to good is that it is catching a panic instead of returning an error. 
image

tarkah:
@meltinglava Thanks for the explanation. The CLI was never intended to be on par with the GUI and was only ever viewed as a means of automating certain tasks. It's true that the `install` command for git was basically a testing ground before we integrated into the GUI, which is why it remains (and not an automation task). Hopefully that gives you context as for why we are using timestamped logging since update / backup commands are intended to be used in scheduled tasks.

I'm not sure the current state of the CLI warrants any fundamental changes, but I do agree it can be improved if we decide to expand on it in the future, especially if other one off tasks get implemented like import / export addon list. 

What do you mean by this line? There is no known place that the program panics (outside the known GUI crashes due to gpu drivers) and everything should safely roll up via Err already.

> Would also like to have more errors return `Result::Err`, rather than panics.

src/main.rs
```diff
@@ -9,33 +9,25 @@ mod gui;
...
     // If this is a clap error, we map to None since we are going to exit and display
     // an error message anyway and this value won't matter. If it's not an error,
     // the underlying `command` will drive this variable. If a `command` is passed
     // on the command line, Ajour functions as a CLI instead of launching the GUI.
-    let is_cli = opts_result
-        .as_ref()
-        .map(|o| &o.command)
-        .unwrap_or(&None)
-        .is_some();
+    let is_cli = opts_result.as_ref().map(|o| &o.command).is_ok();
```
tarkah, line 30: We're not trying to evaluate if the CLI input was properly deserialized (the Result), we are trying to evaluate if a Command was passed on the CLI (the underlying Option). This is needed inside the `validate_opts_or_exit` to properly implement a workaround for windows to attach to the parent console when built / run in release mode.

We need to evaluate this before evaluating if `opts_result` is an error so we can implement the workaround for windows prior to printing the error message to the command line so it'll show up.

meltinglava:
I was wrong here. Just did not suspect the backtrace calls as its usually not needed. (thought they came from panics)

src/main.rs
```diff
@@ -121,36 +114,32 @@ fn setup_logger(is_cli: bool, is_debug: bool) -> Result<()> {
...
+    let log_file = std::fs::OpenOptions::new()
+        .write(true)
+        .create(true)
+        .append(false)
+        .truncate(true)
+        .open(config_dir.join("ajour.log"))?;
 
-        logger = logger.chain(log_file);
-    };
+    logger = logger.chain(log_file);
```
tarkah, line 130: This needs to be reverted. It was working as intended, where we only log to file if not a CLI command and run in release mode. And to stdout if in debug mode or CLI command.

We can change the format for logging when `is_cli` if we want it to look similar to printing directly to std_out, but this gives us way more flexibility in the future if we want to add things like a `verbosity` flag for CLI output by using the stdout chain.

tarkah (changes requested):

tarkah:
> 
> 
> I was wrong here. Just did not suspect the backtrace calls as its usually not needed. (thought they came from panics)

I think it's because you are throwing the error as the return to the top level `main` function. Are you able to print that error to std_err and then exit with an `Ok(()`?

meltinglava:
Think there is some configuration that can be done with color-eyre to make it not report that.

src/main.rs
meltinglava: Fixed in: f97dc59011407aa855192ac0ace0f61a8ada0521

tarkah:
Closing for now as we don't need any big CLI changes currently. We can revisit this if we expand CLI usage in the future or make it more on-par feature-wise with GUI.

PR #456, opened 2020-12-25
feat: delete saved variables

### LINKED ISSUE #7: Delete Config (SavedVariables)
As a User, I want to be able to delete the config (SavedVariables) of an installed addon.

Clicking the Delete Config button should remove all the SavedVariables associated with the addon, and all the addons that is related to it via dependencies, but leaving the addons themselves intact.

Pseudo code:
* User clicks the Delete Config button.
* Determine addons with dependencies to the addon for which config is being deleted.
* (Optionally) Present the user with list of dependent addons and allow the user to cancel the delete config operation.
* Delete SavedVariables of all dependent addons across Account, Realm and Character folders.

casperstorm:
Resolves #7

## Proposed Changes
  - Option in Settings do delete saved variables along side the addon.

## Checklist

- [ ] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/crates/core/src/fs/addon.rs
+++ b/crates/core/src/fs/addon.rs
@@ -33,12 +33,10 @@ pub fn delete_saved_variables(addon_folders: &[AddonFolder], wtf_path: &PathBuf)
             .and_then(|a| a.to_str());
 
         if parent_name == Some("SavedVariables") {
-            // Running `file_stem` twice to catch both `.lua` and `.lua.bak` files.
             let file_name = path
                 .file_stem()
-                .map(|a| Path::new(a))
-                .and_then(|a| a.file_stem())
-                .and_then(|a| a.to_str());
+                .and_then(|a| a.to_str())
+                .map(|a| a.trim_end_matches(".bak"));
 
             // NOTE: Will reject "Foobar_<invalid utf8>".
             if let Some(file_name_str) = file_name {
--- a/crates/core/src/fs/addon.rs
+++ b/crates/core/src/fs/addon.rs
@@ -5,7 +5,7 @@ use crate::{
 };
 use std::collections::HashSet;
 use std::fs::{remove_dir_all, remove_file};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use walkdir::WalkDir;
 
 /// Deletes an Addon and all dependencies from disk.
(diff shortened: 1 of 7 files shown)


casperstorm:
I am wondering if we should just loop through all accounts and delete every found variables.
feedback @tarkah?

casperstorm:
Take a look when you have time @tarkah :)

tarkah:
Will do!

crates/core/src/fs/addon.rs
```diff
@@ -19,6 +20,35 @@ pub fn delete_addons(addon_folders: &[AddonFolder]) -> Result<()> {
...
 }
 
+/// Deletes all saved varaible files correlating to `[AddonFolder]`.
+pub fn delete_saved_variables(addon_folders: &[AddonFolder], wtf_path: &PathBuf) -> Result<()> {
+    for entry in WalkDir::new(&wtf_path)
+        .into_iter()
+        .filter_map(std::result::Result::ok)
+    {
+        let path = entry.path();
+        let file_name = path
```
tarkah, line 30: Should we call this `parent_name` instead?

crates/core/src/fs/addon.rs
```diff
@@ -19,6 +20,35 @@ pub fn delete_addons(addon_folders: &[AddonFolder]) -> Result<()> {
...
+            .and_then(|a| a.file_name())
+            .and_then(|a| a.to_str());
+
+        if file_name == Some("SavedVariables") {
+            let file_name = path.file_name().and_then(|f| f.to_str());
+
+            // NOTE: Will reject "Foobar_<invalid utf8>".
+            if let Some(file_name_str) = file_name {
+                for folder in addon_folders {
+                    if file_name_str.contains(&folder.id) {
```
tarkah, line 41: i haven't researched, but shouldn't the .lua file be named identically to the addon folder? The contains seems dangerous as it might remove things from similarly named addons that weren't intended to be deleted.

tarkah (changes requested):

crates/core/src/fs/addon.rs
casperstorm: You are right. Will fix this.

crates/core/src/fs/addon.rs
```diff
@@ -19,6 +20,40 @@ pub fn delete_addons(addon_folders: &[AddonFolder]) -> Result<()> {
...
+            .and_then(|a| a.file_name())
+            .and_then(|a| a.to_str());
+
+        if parent_name == Some("SavedVariables") {
+            // Running `file_stem` twice to catch both `.lua` and `.lua.bak` files.
+            let file_name = path
+                .file_stem()
+                .map(|a| Path::new(a))
+                .and_then(|a| a.file_stem())
+                .and_then(|a| a.to_str());
```
tarkah, line 41: I think the following would read a little better:

```rust
            let file_name = path
                .file_stem()
                .and_then(|a| a.to_str())
                .map(|a| a.trim_end_matches(".bak"));
```

tarkah (approved):
Looks good. Haven't tested but seems like it'll work.

PR #463, opened 2021-01-01
chore: removed force install from addons

casperstorm:
Resolves better UX.

## Proposed Changes
  - `Force install` had little to no value.

## Checklist

- [ ] Tested on Windows
- [X] Tested on MacOS
- [ ] Tested on Linux


tarkah:
Makes sense! Don't think I've used this since we have channel switching working much more reliably now.

tarkah (approved):

PR #464, opened 2021-01-01
fix: catalog pixel difference

casperstorm:
Resolves a little pixel height difference in catalog. See blow.

Screenshot 2021-01-01 at 15 10 31

## Checklist

- [ ] Tested on Windows
- [X] Tested on MacOS
- [ ] Tested on Linux


tarkah:
I love this so much

tarkah (approved):

PR #473, opened 2021-01-05
feat: localizations

casperstorm:
This PR will introduce the ability to switch language in Ajour.

## Checklist

- [x] Tested on Windows
- [x] Tested on MacOS
- [x] Tested on Linux


src/gui/element/settings.rs
```diff
@@ -30,14 +31,26 @@ pub fn data_container<'a, 'b>(
...
+    let ctx = static_json_gettext_build!(
+        "en_US",
+        "en_US",
+        "locale/en_US.json",
+        "da_DK",
+        "locale/da_DK.json"
+    )
+    .unwrap();
+    let val = localization_state.languages.get(&config.language).unwrap();
+    let a = get_text!(ctx, val, "hello").unwrap();
```
casperstorm, line 50: This is the core of the way we will use this system, @tarkah.
It works simply and nice, imho.

However I need to find a proper place to generate that `ctx` which I need to pass into `get_text!` macro everywhere I use it. Any idea of something nice?

tarkah: Super simple! This is awesome. Can we store the JSONGetText (ctx) struct in Ajour? We can build it / store it during  Application::new ?

src/gui/element/catalog.rs
```diff
@@ -302,6 +313,8 @@ pub fn data_row_container<'a, 'b>(
         })
         .next()
     {
+        // TODO (casperstorm): localization timeago.
+        // @see: https://docs.rs/timeago/0.2.1/timeago/
```
casperstorm, line 317: @tarkah This should properly also be localized, and they have support for it. If you have time you are very welcome to help.

tarkah (approved):
Check out my commits, made some changes. Looks great! Hard to test until we have the danish file fully translated.

PR #475, opened 2021-01-06
Feat/reimplement changelog

### LINKED ISSUE #474: Still on 0.5.3 because of this...
Please bring back the ability to view changelogs directly inside Ajour by clicking the remote version numbers as it used to work before version 0.5.4.

I don't understand why this was removed to begin with. I really liked this feature as it makes it a lot easier to keep up with changelogs rather than having to open up a browser tab for each addon, which in my case can be up to 10 tabs sometimes.

I really don't want to update Ajour until this feature gets re-implemented. 

Anyway, even if this doesn't get re-implemented into Ajour, thanks for this great addon manager!

tarkah:
Resolves #474

## Proposed Changes
  - Add back inline changelog

## Checklist

- [x] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ and `Removed`.
 
 ## [Unreleased]
 
+### Added
+
+- Added back inline changelogs for remote version. Clicking on the remote version
+  will show the changelog inline instead of opening a browser window.
+
 ### Fixed
 
 - Parsing error causing WeakAuras to fail parsing due to missing "version" field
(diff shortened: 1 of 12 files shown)


casperstorm (changes requested):
Great work, quick execution @tarkah.
I think its great, except that it seems like you have to "double click" to exit out of a expanded addon now because you loose the selection once you go into changelog. Thats the only thing i found.

I think @ZythDr will be happy with this PR :)

tarkah:
Good catch! I'll make sure the expanded row closes on first click, regardless of if the addon details are expanded or changelog is expanded.

PR #480, opened 2021-01-09
feat: relative time localization

casperstorm:
`timeago` crate had support for localization, i've implemented it with fallback to english.

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -65,6 +65,7 @@ dependencies = [
  "iced_native",
  "image",
  "isahc",
+ "isolang",
  "json-gettext",
  "log",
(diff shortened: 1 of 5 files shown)


tarkah:
Looks spot on!

Nelfym:
Thanks from France 😁.
I assume you'll add localization keys later?

casperstorm:
> Thanks from France 😁.
> I assume you'll add localization keys later?

French is [supported](https://github.com/vi/timeago/blob/master/src/languages/french.rs).

PR #496, opened 2021-01-18
feat: additional languages

casperstorm:
## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux


src/localization.rs
```diff
@@ -40,7 +40,12 @@ pub fn localized_string(key: &str) -> String {
     let lang = LANG.get().expect("LANG not set").read().unwrap();
 
     if let Some(text) = get_text!(LOCALIZATION_CTX, *lang, key) {
-        text.to_string()
+        let text = text.to_string();
+        if text.is_empty() {
+            key.to_owned()
+        } else {
+            text
+        }
```
casperstorm, line 48: It's not so pretty, @tarkah but how do you like this?
The idea is that if we have a language missing a translation we often create it with `""` which would then at least return the key.

tarkah: I think that makes a lot of sense. Will be helpful when adding a new language to see where each key is in the gui for additional context

PR #497, opened 2021-01-18
fuzzy match tweaks

### LINKED ISSUE #462: Better sorting of catalog search results
When searching the catalog, changing the text in the search field resets the sort order to unsorted

I would expect the selected column sorting to be applied to the search results and not the default order.

I can't think of any other than reapplying the sort order repeatedly when searching.

**Additional context**
Tested on Ajour 0.6.0 on Mac OS X 10.15.7 with the Curse catalog and the Tukui catalog.

To reproduce:
Search for "Deadly Boss Mods (DBM)"
Change the sort order to "Game Version" descending
Add a " " (space character) to the end of the search string
Notice that the column sorting is removed and a default order is applied.

More severe case:
Search for "Adibags"...

tarkah:
Resolves #462

## Proposed Changes
  - Filter out scores that are negative. When I first implemented this, I thought it would only return positive scores by default but never checked. This helps remove a ton of non-related results.
  - Increase the penalty for when gaps occur in matches characters vs the search pattern. This helps push even more results negative and cleans up the results. We prefer matches that closely match the search pattern or at least have consecutive characters and not too many spaces between matching characters.

## Checklist

- [x] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and `Removed`.
 ### Changed
 
 - Refreshed the menu with a more simple and uncluttered look.
+- Catalog fuzzy matching better removes unrelated results while retaining relevant
+  results
 
 ## [0.6.3] - 2021-01-14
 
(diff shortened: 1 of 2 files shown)


casperstorm (approved):
Nice one.

PR #503, opened 2021-01-20
en_US spelling

Xarano-GIT:
changed 'diretory' to 'directory'


casperstorm:
Thank you for this PR.
I have corrected it on our localization page. You can find it here: https://poeditor.com/join/project?hash=gjghTR7d0Z.
I would like localization corrections to go through that site because we sync from it :-)

PR #515, opened 2021-01-26
Update se_SE.json

ddund:
I changed language to swedish and saw some words that bothered me so here is my proposal for what they could be replaced with

Line 2: Could also just be "Om"="About". I changed it to "Om Ajour" instead of "Om Appen"="About the app".
Line 8: Alternativa -> Alternerande: alternate != alternativa. In swedish it's called "alternerande"
Line 34: Visa inte -> Dölj: Its a positive mixed with a negative. Better to just use one negative
Line 52: "... för installerade för" looks weird but I don't know how to trigger it in the client so not sure about the context. My guess is that it should be changed to "Inga Addons installerade för ..."
Line 62 "refresh". I think "Ladda om" is kinda weird. The best word would be "Uppdatera" but that might be confusing with "Uppdatera alla" next to it. I think that maybe a better word would be "Hämta uppdateringar" = "Get updates" instead.

Also, would it be possible to have the languages in the drop-down menu sorted?


ddund:
Hey,

I totally missed that you could collaborate on POEditor. I will close this and add my comments there instead.

PR #518, opened 2021-01-27
added: zh_CN

CyanMY:
Resolves #

## Proposed Changes
  -

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux


casperstorm:
Thank you so much for the contribution!
`Iced` however does not support shaping right now which means it will not render hanzi correctly.

So we are awaiting https://github.com/hecrj/iced/issues/33.

CyanMY:
> Thank you so much for the contribution!
> `Iced` however does not support shaping right now which means it will not render hanzi correctly.
> 
> So we are awaiting [hecrj/iced#33](https://github.com/hecrj/iced/issues/33).

OK.
Looking forward to that time will come sooner.

PR #521, opened 2021-01-29
feat: context localizations

casperstorm:
## Proposed Changes
  - Quite a few error has not yet been localized.

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -48,7 +48,7 @@ dependencies = [
 
 [[package]]
 name = "ajour"
-version = "0.7.0"
+version = "0.6.0"
 dependencies = [
  "ajour-core",
  "ajour-weak-auras",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "ajour"
 description = "A World of Warcraft addon manager"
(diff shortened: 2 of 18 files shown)


casperstorm:
@tarkah Would you be able to look at getting a generic "IO error" for the `error-update-ajour` as we talked about on Discord? Im a little unsure how to do it with the context and AnyHow 🤔 I am thinking we can have a `error-io-update-ajour` for it, or just use the other if we only want one?

tarkah:
Sounds good! I'll at least add the code to match the downcasted error on IO error and log things so we can test what actually happens when we try to self update from a folder we don't have permissions to. Then you can play around with an error message for it.

tarkah:
@casperstorm Check out my latest commit, it should make sense how I'm downcasting to get to check for io error / permissions issue. I've not added the "key" for localization yet, figured you could mess with that.

Note I modified the version so we could test the self update feature and I tested by copying to my C:\ root and was able to get the custom context for permissions issue! Make sure to revert version before merging.

casperstorm:
Make sense, and works perfectly!

PR #522, opened 2021-01-29
feat: added option categories column to catalog

casperstorm:
## Proposed Changes
  - Added option categories column to catalog

## Checklist

- [ ] Tested on Windows
- [X] Tested on MacOS
- [ ] Tested on Linux


tarkah (approved):
Looks and tests good!

PR #523, opened 2021-01-29
tweak search settings

### LINKED ISSUE #517: Search broken?
The search function in 0.7 seems to be broken. For example, searching for "minimap" fails to show some minimap-related addons, including some that have "Minimap" right in the name! e.g. BasicMinimap.

tarkah:
Resolves #517 

## Proposed Changes
  - Score title and description separately, with a higher weighting to title. Use the max of those 2 scores.
  - This should help sort the matches better on title first, which is more relevant generally to what the user wants to see

## Checklist

- [x] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@ and `Removed`.
 - Optional 'Categories' column for Catalog.
 - Optional 'Summary' column for My Addons.
 
+## Changed
+
+- Tweaked catalog fuzzy search to weight title matches higher than description
+  matches
+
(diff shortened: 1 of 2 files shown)


casperstorm:
This looks good to me if you want it into deveopment. Maybe we should add a note to changelog?

casperstorm:
Should we merge this, @tarkah?

tarkah:
@casperstorm updated CL and merged!

PR #529, opened 2021-02-06
add support for townlong-yak addons

### LINKED ISSUE #512: TownLong-Yak Addons Repository
The TownLong-Yak addons (OPie, etc.) aren't kept up to date in any of the standard repositories (OPie has been several versions behind for months now, and is missing bug fixes and a lot of Shadowlands functionality).

These addons are quite popular, TownLong-Yak is a pretty established WoW resource, and they ought to be added to Ajour as a repository.

The site has an API to check for updates: https://www.townlong-yak.com/addons/about/update-api

tarkah:
Resolves #512 

## Proposed Changes
  - Add townlong-yak addon support via the hub api

## Checklist

- [x] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/crates/core/src/cache.rs
+++ b/crates/core/src/cache.rs
@@ -195,6 +195,7 @@ pub struct AddonCacheEntry {
     pub primary_folder_id: String,
     pub folder_names: Vec<String>,
     pub modified: DateTime<Utc>,
+    pub external_release_id: Option<ExternalReleaseId>,
 }
 
 impl TryFrom<&Addon> for AddonCacheEntry {
@@ -207,6 +208,15 @@ impl TryFrom<&Addon> for AddonCacheEntry {
             let mut folder_names: Vec<_> = addon.folders.iter().map(|a| a.id.clone()).collect();
             folder_names.sort();
 
+            let external_release_id = if repository == RepositoryKind::TownlongYak {
+                addon.file_id().map(ExternalReleaseId::FileId)
+            } else {
+                addon
+                    .version()
+                    .map(str::to_string)
+                    .map(ExternalReleaseId::Version)
+            };
+
             Ok(AddonCacheEntry {
                 title: addon.title().to_owned(),
                 repository,
@@ -214,6 +224,7 @@ impl TryFrom<&Addon> for AddonCacheEntry {
                 primary_folder_id: addon.primary_folder_id.clone(),
                 folder_names,
                 modified: Utc::now(),
+                external_release_id,
             })
         } else {
             Err(CacheError::AddonMissingRepo {
@@ -221,6 +232,12 @@ impl TryFrom<&Addon> for AddonCacheEntry {
             })
         }
     }
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub enum ExternalReleaseId {
+    FileId(i64),
+    Version(String),
 }
 
 #[derive(Serialize, Deserialize, Debug)]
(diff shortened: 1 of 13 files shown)


crates/core/src/cache.rs
```diff
@@ -207,13 +208,23 @@ impl TryFrom<&Addon> for AddonCacheEntry {
...
 
+            let external_release_id = if repository == RepositoryKind::TownlongYak {
+                addon.file_id().map(ExternalReleaseId::FileId)
+            } else {
+                addon
+                    .version()
+                    .map(str::to_string)
+                    .map(ExternalReleaseId::Version)
+            };
+
```
tarkah, line 219: This is how we cache the version that was just installed / updated. When we reload the addon from the cache, it'll use this as the version / file_id to check against the remote API. This resolves issues where the addon .toc version is not comparable to the remote version on the API

crates/core/src/catalog.rs
```diff
@@ -7,7 +7,7 @@ use chrono::prelude::*;
 use isahc::ResponseExt;
 use serde::{Deserialize, Serialize};
 
-const CATALOG_URL: &str = "https://github.com/ajour/ajour-catalog/raw/master/catalog.json";
+const CATALOG_URL: &str = "https://raw.githubusercontent.com/ajour/ajour-catalog/3ce1414498adbc7167f11fe0dac7cd58e7b55391/catalog.json";
```
tarkah, line 10: Make sure we change this back before merging. I switched over to test.

crates/core/src/parse.rs
```diff
@@ -688,7 +731,20 @@ fn build_addons(
...
+            // validate the installed version against the remote version
+            match &e.external_release_id {
+                Some(ExternalReleaseId::FileId(file_id)) => {
+                    repo_package.metadata.file_id = Some(*file_id)
+                }
+                Some(ExternalReleaseId::Version(version)) => {
+                    repo_package.metadata.version = Some(version.clone())
+                }
+                None => {}
+            }
```
tarkah, line 747: Here is where we apply that cached remote release version when building addon from cache entry

src/gui/update.rs
```diff
@@ -757,15 +757,16 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
...
-                    version = Some(package.version);
-                }
-                if let Some(version) = version {
-                    addon.set_version(version);
+                    addon.set_version(package.version);
+
+                    if let Some(file_id) = package.file_id {
+                        addon.set_file_id(file_id);
+                    }
                 }
```
tarkah, line 767: Here is where we set version / file id from the newly installed / updated package. This is done prior to updating the cache entry, so cache will always stay up to date with latest external release id when installing / updating

src/gui/mod.rs
```diff
@@ -1838,6 +1844,7 @@ impl std::fmt::Display for CatalogSource {
                 catalog::Source::Curse => "Curse",
                 catalog::Source::Tukui => "Tukui",
                 catalog::Source::WowI => "WowInterface",
+                catalog::Source::TownlongYak => "TownlongYak",
```
casperstorm, line 1847: Wondering if we should write: Townlong Yak

casperstorm:
I've tested this on both Windows and macOS and it works perfectly.
I don't really have any comments  to the code. Its clean. Good job, once again mate.

src/gui/mod.rs
tarkah: Might as well put a `-` in at that point?

casperstorm: Yeah lets do that.

PR #530, opened 2021-02-07
Release 0.7.1

casperstorm:
This is the PR for release 0.7.1.

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux


casperstorm:
@tarkah running with:

```
const CATALOG_URL: &str = "https://raw.githubusercontent.com/ajour/ajour-catalog/3ce1414498adbc7167f11fe0dac7cd58e7b55391/catalog.json";
```

On `master` will cause the following problem:

```sh
09:01:19.766 [ajour][ERROR] caused by: unknown variant `townlong-yak`, expected one of `Curse`, `Tukui`, `WowI` at line 1 column 5838769
```

A little unsure how we do this migration the best.

tarkah:
Ahh yes 🤔 too bad we didn't have the foresight to put in a catch all variant for unknown values.

We probably should publish a v2 catalog file to the repo and pick that one up. And add that catch all unknown variant so we don't have this issue again in the future.

casperstorm:
> Ahh yes 🤔 too bad we didn't have the foresight to put in a catch all variant for unknown values.
> 
> We probably should publish a v2 catalog file to the repo and pick that one up. And add that catch all unknown variant so we don't have this issue again in the future.

Agree.
Is it something you could do in client? Then I'll take a look what we can do in catalog.

casperstorm:
Take a look at the new catalog, @tarkah.

PR #531, opened 2021-02-07
fix: format townlong-yak game versions

casperstorm:
Formats the townlong-yak game versions from catalog into the what we use the other places.

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/src/gui/element/catalog.rs
+++ b/src/gui/element/catalog.rs
@@ -6,7 +6,8 @@ use {
     },
     crate::localization::{localized_string, localized_timeago_formatter},
     ajour_core::{
-        config::Config, theme::ColorPalette, utility::format_interface_into_game_version,
+        catalog::Source, config::Config, theme::ColorPalette,
+        utility::format_interface_into_game_version,
     },
     ajour_widgets::{header, Header, TableRow},
     chrono::prelude::*,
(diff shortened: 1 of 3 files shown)


src/gui/element/catalog.rs
```diff
@@ -276,7 +278,7 @@ pub fn data_row_container<'a, 'b>(
             .game_versions
             .iter()
             .find(|gv| gv.flavor == config.wow.flavor.base_flavor())
-            .map(|gv| gv.game_version.clone())
+            .map(|gv| format_interface_into_game_version(&gv.game_version[..]))
```
tarkah, line 281: All other backends have a properly formatted game version that isn't 5 characters? Just want to make sure this couldn't ever impact data from the other backends?

casperstorm: I've added test to at least ensure it's as I expect. But don't know if it's prettier we do it earlier in the deser progress

tarkah: I think it's fine to leave it here... maybe just add logic to only evaluate that mapping for TownlongYak source? `if addon_data.source == TownlongYak { format_interface_into_game_version(&gv.game_version[..]) } else { gv.game_version.clone() }`

casperstorm: I think that make sense to do. Ill have to go to bed so the RC has to wait. If you dont add it ill do it tomorrow tomorrow :-)

tarkah: Sounds good, have a good night!

PR #532, opened 2021-02-08
fix: catalog source fallback

casperstorm:
This will resolve the us being able to keep adding sources to the catalog without breaking as just happened.

## Checklist

- [ ] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/src/gui/mod.rs
+++ b/src/gui/mod.rs
@@ -1845,6 +1845,7 @@ impl std::fmt::Display for CatalogSource {
                 catalog::Source::Tukui => "Tukui",
                 catalog::Source::WowI => "WowInterface",
                 catalog::Source::TownlongYak => "Townlong-Yak",
+                catalog::Source::Other => panic!("Unspported catalog source"),
             },
         };
         write!(f, "{}", s)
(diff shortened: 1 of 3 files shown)


src/gui/mod.rs
```diff
@@ -1845,6 +1845,7 @@ impl std::fmt::Display for CatalogSource {
                 catalog::Source::Tukui => "Tukui",
                 catalog::Source::WowI => "WowInterface",
                 catalog::Source::TownlongYak => "Townlong-Yak",
+                catalog::Source::Other => panic!("Unspported catalog source"),
```
casperstorm, line 1848: Should be fine to `panic!` in the `std::fmt::Display` since we don't want to render this to the GUI.

src/gui/update.rs
```diff
@@ -2134,6 +2134,7 @@ async fn perform_fetch_latest_addon(
                     catalog::Source::Tukui => RepositoryKind::Tukui,
                     catalog::Source::WowI => RepositoryKind::WowI,
                     catalog::Source::TownlongYak => RepositoryKind::TownlongYak,
+                    catalog::Source::Other => panic!("Unspported catalog source"),
```
casperstorm, line 2137: Should be fine to `panic!` since it shouldn't be possible to end here with a `catalog::Source::Other`.

casperstorm:
Care to take a look at this @tarkah, and see if this was your idea as well.

src/gui/mod.rs
tarkah: Yup! The only way it'll expose itself in the GUI and make allow any catalog items for it to display is if you add it to the CatalogSource::all method. Since we left it out, these should be unreachable and it'd be impossible to show those items in the GUI.

tarkah (approved):

PR #544, opened 2021-02-18
fix: update cache when download townlong addons

casperstorm:
Resolves: 
Running CLI update when theres updates for TownlongYak addons kept saying they could be updated.
It seems to be because we dont update the cache. Can you confirm this, @tarkah?

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/src/command/update_addons.rs
+++ b/src/command/update_addons.rs
@@ -208,6 +208,15 @@ async fn update_addon(
         }
     }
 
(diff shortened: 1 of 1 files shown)


tarkah (approved):
Works like a charm! I had an update for venture plan and updating via CLI updated the cache appropriately. When I reran `update-addons` via CLI, no updates were available.

PR #545, opened 2021-02-19
fix: wording in weakauras

casperstorm:
Resolves:
I don't like the way we worded WeakAura updates. I think we should move to "Completed" just like addons and then still leave the  "Finish ingame" message we have:

weakaura_fix

## Checklist

- [x] Update terms on POEditor (remove weakaura-update-queued)
- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/src/localization.rs
+++ b/src/localization.rs
@@ -184,6 +184,7 @@ mod test {
         update_all: String,
         ajour_update_channel: String,
         updating: String,
+        weakaura_update_available: String,
         weakaura_updates_queued: String,
         weakauras_loaded: String,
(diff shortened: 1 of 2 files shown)


src/localization.rs
```diff
@@ -184,6 +184,7 @@ mod test {
         update_all: String,
         ajour_update_channel: String,
         updating: String,
+        weakaura_update_available: String,
```
casperstorm, line 187: This was missing, has nothing to do with this PR.

PR #546, opened 2021-02-21
For users with only a single account, automatically pick the account when in the weak aura view

nkf:
Resolves #

## Proposed Changes
  - For users with only a single account, automatically pick the account when in the weak aura view

## Checklist

- [x] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/src/gui/update.rs
+++ b/src/gui/update.rs
@@ -1875,21 +1875,23 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
                 let state = ajour.weak_auras_state.entry(flavor).or_default();
                 state.accounts = accounts;
 
-                let single_account = if state.accounts.len() == 1 {
-                    Some(state.accounts[0].clone())
-                } else {
-                    None
-                };
-
                 // If we have an account already selected, use that as the picklist selection
                 // or if there is only a single account to choose, use that
                 // and trigger a parse for this without user interaction
-                if let Some(account) = ajour.config.weak_auras_account.get(&flavor).or(single_account.as_ref()) {
+                let account_from_config = ajour.config.weak_auras_account.get(&flavor).cloned();
+                let get_single_account = || {
+                    if state.accounts.len() == 1 {
+                        Some(state.accounts[0].clone())
+                    } else {
+                        None
+                    }
+                };
+                if let Some(account) = account_from_config.or_else(get_single_account) {
                     if let Some(wtf_path) = ajour.config.get_wtf_directory_for_flavor(&flavor) {
                         state.chosen_account = Some(account.clone());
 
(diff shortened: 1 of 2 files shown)


src/gui/update.rs
```diff
@@ -1876,13 +1876,22 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
                 state.accounts = accounts;
 
                 // If we have an account already selected, use that as the picklist selection
+                // or if there is only a single account to choose, use that
                 // and trigger a parse for this without user interaction
-                if let Some(account) = ajour.config.weak_auras_account.get(&flavor) {
+                let account_from_config = ajour.config.weak_auras_account.get(&flavor).cloned();
+                let get_single_account = || {
+                    if state.accounts.len() == 1 {
+                        Some(state.accounts[0].clone())
```
casperstorm, line 1884: Maybe a little prettier to use `.first()` (https://doc.rust-lang.org/std/vec/struct.Vec.html#method.first)?

casperstorm (changes requested):
Great addition, i really love these QOL changes. Also welcome to the contributor team. It's nice to see new people help out on Ajour!

src/gui/update.rs
nkf: `.first()` here is a bit awkward since it returns an Option.

It would be convenient if there existed an 'exactly one' function that returned the first element if it was the only element.

Writing this made me realise that maybe this is a job for pattern matching

casperstorm: Yeah, thats true. Otherwise I am fine with using this for now.

casperstorm (approved):

nkf:
Nice!

Thanks for creating this awesome tool.

Do you use rust-analyzer for this by any chance? I couldn't get it to work for this project, it just never completes the initial analysis

casperstorm:
> 
> 
> Nice!
> 
> Thanks for creating this awesome tool.
> 
> Do you use rust-analyzer for this by any chance? I couldn't get it to work for this project, it just never completes the initial analysis

I use rust-analyzer! However @tarkah has reported problems on his end with it as well. 🤔

tarkah:
Yeah I have to use a releases that are ~3+ months old, otherwise it always fails when opening gui/update.rs. It's very annoying.

tarkah:
Also, great addition! I love the pattern match on a single item slice

casperstorm:
@nkf can i get you to update CHANGELOG.md?

nkf:
> 
> 
> @nkf can i get you to update CHANGELOG.md?

Sure, I wasn't sure if this was too trivial to go into the changelog, ill update the log :)

PR #548, opened 2021-02-22
chore: update dependencies

casperstorm:
Just updating a few dependencies.

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux
- [ ] ~Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users~


tarkah:
@casperstorm I made the breaking changes updates for `isahc`. They have proper async support now for the json de-serialization!

PR #549, opened 2021-02-22
remove old addon folders when updating

tarkah:
## Proposed Changes
  - When updating an addon, make sure all folders it had _before_ the update are deleted. This helps prevent orphaned folders from being created, which happens notably with addons like DBM.

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux


casperstorm:
Lets make sure to test this properly when RC goes out.

PR #550, opened 2021-02-22
Fix/disable self update linux nonappimage

tarkah:
## Proposed Changes
  - Self update for Linux only works for the published AppImage. Disable the update button for non-appimage binaries.

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ and `Removed`.
 ### Fixed
 
 - Issue with TownlongYak addons while updating them through CLI
+- Linux: Disable self update for non-AppImage binaries since this functionality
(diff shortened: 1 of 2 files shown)


tarkah:
@casperstorm This isn't tested but it's super straightforward and I'm pretty positive it has the intended effects. The `needs_update` variable determines whether we show the Update button in place of the About button. I ALWAYS set that true if we're not on linux. If we are on linux, I only set it true if the APPIMAGE env variable is set (always set when running an appimage binary).

casperstorm:
Seems good to me too, great work.

PR #553, opened 2021-02-24
feat: ux update catalog

casperstorm:
Resolves ux problems with people not finding sources in the catalog.
Note this is not done yet.

Screenshot 2021-02-24 at 16 54 30

## Proposed Changes
  -

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/src/gui/update.rs
+++ b/src/gui/update.rs
@@ -1573,13 +1573,11 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
             log::debug!("Interaction::CatalogResultSizeSelected({:?})", source);
 
             // Catalog source
-            match source {
-                CatalogSource::Choice(source) => {
+            if let CatalogSource::Choice(source) = source {
+
                     ajour.config.catalog_source = Some(source);
                     let _ = ajour.config.save();
-                }
-                _ => {}
-            };
+            }
 
             ajour.catalog_search_state.categories = ajour
                 .catalog_categories_per_source_cache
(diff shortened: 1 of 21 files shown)


casperstorm:
I'm not far, but @tarkah take a look at what you think about the solution.

tarkah:
@casperstorm I don't think we need the unnecessary "Empty" catalog source.

If we add a new variant to `CatalogSource` like

```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
pub enum CatalogSource {
    Choice(catalog::Source),
    None, // New variant
}
```

And then make the config one an option:

```rust
    #[serde(default)]
    pub catalog_source: Option<catalog::Source>,
```

Then we can assign the None variant when the config option isn't set (None) here:

old 
```rust
let catalog_source = CatalogSource::Choice(self.config.catalog_source);
```

new
```rust
let catalog_source = self.config.catalog_source.map(CatalogSource::Choice).unwrap_or(CatalogSource::None);
```

casperstorm:
Great feedback as always, @tarkah.
Take a look again now.

src/gui/update.rs
```diff
@@ -1442,10 +1442,15 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
             });
 
             ajour.catalog_categories_per_source_cache = categories_per_source;
+            let catalog_source_choice = if let Some(source) = &ajour.config.catalog_source {
```
casperstorm, line 1445: Not sure if this could be done prettier with a `.map()`.

tarkah:
```rust
ajour
.config
.catalog_source
.map(CatalogSource::Choice)
.unwrap_or(CatalogSource::None)
```
?

tarkah (approved):
Looks good, just needs a quick rustfmt run on it!

PR #562, opened 2021-03-12
feat/auto-update

tarkah:
## Proposed Changes
  - Add feature to automatically check for new updates on interval
  - Also add option to enable applying new updates automatically

## Checklist

- [ ] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux


tarkah:
@casperstorm Test this out. I've currently only implemented "checking" for new updates. It's hard-coded to check every 30 minutes. Should we make that configurable or is 30mins sane? Considering how slow to update Curse API is, not sure that matters too much.

Adding the ability to apply these updates automatically will be trivial. I'll make sure that's opt-in via config.

When an update is found for the Flavor, it'll resort the addons by status to push those updates to the top of the screen.

src/gui/update.rs
```diff
@@ -527,6 +529,104 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
...
+                                        addon.title(),
+                                        addon.version().unwrap_or_default(),
+                                        package.version
+                                    );
+
+                                    addon.state = AddonState::Updatable;
+
+                                    has_update += 1;
+                                } else {
+                                    addon.state = AddonState::Idle;
```
tarkah, line 605: @casperstorm I'm thinking of removing this line. If an addon is updated, we show the status as "completed" until the user "refreshes". This would effectively overwrite those statuses every X minutes this process runs. What do you think?

casperstorm: I agree, @tarkah. Sounds good.

casperstorm:
This is a great PR, @tarkah. It looks great.

> Should we make that configurable or is 30mins sane? Considering how slow to update Curse API is, not sure that matters too much.

I have thought about it for a while and initially I wanted to make it quite customisable with a drop-down of different intervals, but then again... why? As you also mention we are limited by the API and 30mins should be MORE than enough and would simplify the GUI (and UX) a lot. I suggest we add a checkbox to enable "auto update" and enable auto refresh for everyone.

The auto refresh i feel is a QOL thing everyone would benefit from.

tarkah:
@casperstorm I 100% agree with you. Worst case, we get people asking for control and we can add the interval down the road.

casperstorm:
@tarkah I've removed the test in `localization.rs` because we often forgot to add keys and we don't really need it now since everything comes from poeditor and they handle that all languages has the same terms.

Other than that all is LGTM!

casperstorm (approved):

PR #564, opened 2021-03-12
Feat/plater support

### LINKED ISSUE #552: Ability to update Plater scripts/mods just like weakauras from Wago
The title of the feature request pretty much sums it up.
Add the ability to update Plater scripts and mods from the wago repository, just like weakauras.

tarkah:
Resolves #552 

## Proposed Changes
  - Ability to queue updates for Plater profiles, scripts and hooks from My WeakAuras tab

## Checklist

- [x] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux


tarkah:
@casperstorm this _should_ work, but I need someone to test with MANY profiles / mods imported. I've also not tested the companion addon works. It should,  format looks correct, but I don't have WOW installed anymore to test.

crates/weak_auras/src/lib.rs
```diff
@@ -373,6 +545,27 @@ struct AuraChangelog {
...
+    WeakAura,
+    PlaterProfile,
+    PlaterHook,
+    PlaterScript,
+}
+
+impl Display for AuraDisplayKind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let s = match self {
+            AuraDisplayKind::WeakAura => "Weak Aura",
```
casperstorm, line 559: We use WeakAura across the app, so we should properly also here.

casperstorm:
I have tested, and both Aura, Hook, Script and Profile is updated!!

In Plater you have "Profiles" which is a overall profile (can also include its own scripts/hooks, bundled).
Then you can have seperate script/hooks - an example is "Shorter names" script which modifies just the name of the unitframe.

casperstorm:
One last thing i thought about. I think we should rename the tab from "My WeakAuras" to "My Wagos" or "My Auras".
@mlablah feedback on this?

PR #574, opened 2021-03-16
feat: ability to use multiple wow folders with a single ajour instance

casperstorm:
With this PR it is possible to manage multiple World of Warcraft folders. Use case is if you have retail and classic in two different dirs.

## Checklist

- [x] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux


casperstorm:
@tarkah im pretty happy with the outcome of this PR. Please let me know what you think.

crates/core/src/config/mod.rs
```diff
@@ -64,15 +64,28 @@ pub struct Config {
...
 impl Config {
+    /// Returns a `PathBuf` to the flavor directory.
+    pub fn get_flavor_directory_for_flavor(&self, flavor: &Flavor, path: &PathBuf) -> PathBuf {
+        path.join(&flavor.folder_name())
+    }
+
+    /// Returns a `Option<PathBuf>` to the root directory of the Flavor.
+    pub fn get_root_directory_for_flavor(&self, flavor: &Flavor) -> Option<PathBuf> {
+        if let Some(flavor_dir) = self.wow.directories.get(flavor) {
+            Some(Box::new(flavor_dir.parent().unwrap()).to_path_buf())
```
tarkah, line 75: Is the box necessary here?

crates/core/src/config/mod.rs
```diff
@@ -108,25 +121,19 @@ impl Config {
...
     pub fn get_download_directory_for_flavor(&self, flavor: Flavor) -> Option<PathBuf> {
-        match self.get_addon_directory_for_flavor(&flavor) {
-            Some(dir) => {
-                // The path to the directory which hold the temporary zip archives
-                let dir = dir.parent().expect("Expected Addons folder has a parent.");
-                Some(dir.to_path_buf())
-            }
-            None => None,
-        }
+        self.wow.directories.get(&flavor).cloned()
```
tarkah, line 126: Before we used the `Interface` folder, but this is using the `\_flavor\_` directory now?

src/gui/update.rs
```diff
@@ -70,21 +70,29 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
...
+                    if flavor_path.exists() {
+                        ajour.config.wow.directories.insert(*flavor, flavor_path);
+                    }
+                }
+
+                ajour.config.wow.directory = None;
+                let _ = &ajour.config.save();
+            }
+
+            let flavors = ajour.config.wow.directories.keys().collect::<Vec<_>>();
```
tarkah, line 88: Should we move this to our `apply_config` function in `src/gui` since its run only once, immediately on startup before the GUI loads the first time?

src/gui/update.rs
```diff
@@ -263,19 +278,39 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
...
+                log::debug!("Message::UpdateWowDirectory(Resolution({:?}))", &path);
+
+                if let Some(flavor) = flavor {
+                    // If a flavor is supplied we only update path for that specific flavor.
+                    let flavor_path = ajour.config.get_flavor_directory_for_flavor(&flavor, &path);
+                    if flavor_path.exists() {
+                        ajour.config.wow.directories.insert(flavor, flavor_path);
+                    }
+                } else {
+                    // If no flavor is supplied it will find as many flavors as possible in the path.
```
tarkah, line 297: If I understand correctly, the only time `None` is passed is when the user has NO directories selected (first time launching Ajour)? Otherwise, it always for a single flavor? Smart.

tarkah:
@casperstorm looks good. I've added some comments but it's small stuff. I need to do some testing still. Awesome job!

crates/core/src/config/mod.rs
casperstorm: Mainly to [convert Path to PathBuf](https://doc.rust-lang.org/std/path/struct.Path.html#see-also-2). Do you want to do it in another way?

crates/core/src/config/mod.rs
casperstorm: I just thought this was easier, honestly.
I am happy to change it if you think it makes the most sense?

src/gui/update.rs
casperstorm: Good point, ill look at that.

src/gui/update.rs
casperstorm: Yeah exactly - I think it makes the most sense that we do a "find all flavors initially from this folder" and from that point we only work on flavor base.

crates/core/src/config/mod.rs
tarkah: Path already has a function `to_path_buf`, the box is unnecessary https://doc.rust-lang.org/std/path/struct.Path.html#method.to_path_buf

crates/core/src/config/mod.rs
tarkah: Probably doesn't matter either way as we clean up those temp files. Your call.

src/gui/update.rs
tarkah: Great! And as we discussed, probably worth keeping some logic in the Message::Parse where it applies any found new flavor folders to unmapped flavor directories, in instances where people install classic to the existing wow directory or something

casperstorm: I moved it to `apply_config` - how do you like it there? https://github.com/ajour/ajour/pull/574/commits/2285ed23d91203b01522ba9e3b74b67be0dcf62f

casperstorm:
How and where should we check for newly added paths? One way is to do the following:

```rust
    // Check if any new flavor has been added since last time.
    // Get missing flavors.
    let mut missing_flavors: Vec<&Flavor> = vec![];
    for flavor in Flavor::ALL.iter() {
        if let None = ajour.config.wow.directories.get(flavor) {
            missing_flavors.push(flavor);
        }
    }

    let flavors = &ajour.config.wow.directories.keys().collect::<Vec<_>>();
    for flavor in flavors {
        // Find root dir of the flavor and check if any of the missing_flavor's is there.
        // If it is, we added it to the directories.
        if let Some(root_dir) = ajour.config.get_root_directory_for_flavor(flavor) {
            for missing_flavor in &missing_flavors {
                let flavor_dir = ajour
                    .config
                    .get_flavor_directory_for_flavor(missing_flavor, &root_dir);
                if flavor_dir.exists() {
                    ajour
                         .config
                         .wow
                         .directories
                         .insert(**missing_flavor, flavor_dir);
                }
            }
        }
    }
```

tarkah:
> 
> 
> How and where should we check for newly added paths? One way is to do the following:
> 
> ```rust
>     // Check if any new flavor has been added since last time.
>     // Get missing flavors.
>     let mut missing_flavors: Vec<&Flavor> = vec![];
>     for flavor in Flavor::ALL.iter() {
>         if let None = ajour.config.wow.directories.get(flavor) {
>             missing_flavors.push(flavor);
>         }
>     }
> 
>     let flavors = &ajour.config.wow.directories.keys().collect::<Vec<_>>();
>     for flavor in flavors {
>         // Find root dir of the flavor and check if any of the missing_flavor's is there.
>         // If it is, we added it to the directories.
>         if let Some(root_dir) = ajour.config.get_root_directory_for_flavor(flavor) {
>             for missing_flavor in &missing_flavors {
>                 let flavor_dir = ajour
>                     .config
>                     .get_flavor_directory_for_flavor(missing_flavor, &root_dir);
>                 if flavor_dir.exists() {
>                     ajour
>                          .config
>                          .wow
>                          .directories
>                          .insert(**missing_flavor, flavor_dir);
>                 }
>             }
>         }
>     }
> ```

Agreed, lets put this under the Parse message so it checks everytime the user launches or refreshes

casperstorm:
@tarkah i've added the logic we discussed now. I found another case I solved as well. As i see it this is the QA:

- [ ] You should **not** see a "Welcome to Ajour" when coming from a previous version. Check settings to see if mapping was successful. 
- [ ] Adding a new path to a flavor you don't have works.
- [ ] You can see addons in the new path.
- [ ] Adding a new path, to an already existing flavor should overwrite it.
- [ ] You can see addons in the new (overwritten) path.
- [ ] Removing a known flavor path should remove it during refresh.
- [ ] Adding a new flavor to a known path should add it automatically.

PR #584, opened 2021-03-22
chore: release_to_draft cleanup

casperstorm:
Resolves:
We always write `Ajour {tag}` so i figured we should add it to the github action.


casperstorm:
@mlablah can you review this just to be sure?

casperstorm:
From now on i'll leave tag as release title as well. Closing this.

PR #588, opened 2021-03-24
feat: classic beta flavor

casperstorm:
The TBC classic beta is using `_classic_beta_` flavor which is new.

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -90,7 +90,7 @@ pub enum Command {
     UpdateWeakauras,
     /// Install an addon from the command line
     Install {
-        #[structopt(parse(try_from_str = str_to_flavor), possible_values = &["retail","ptr","beta","classic","classic_ptr"])]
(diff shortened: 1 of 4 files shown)


src/cli.rs
```diff
@@ -90,7 +90,7 @@ pub enum Command {
     UpdateWeakauras,
     /// Install an addon from the command line
     Install {
-        #[structopt(parse(try_from_str = str_to_flavor), possible_values = &["retail","ptr","beta","classic","classic_ptr"])]
+        #[structopt(parse(try_from_str = str_to_flavor), possible_values = &["retail","ptr","beta","classic","classic_ptr","classic_beta"])]
```
casperstorm, line 93: @tarkah i was wondering if we could take this and move it to `config/wow` so we could have a method on Flavor which returned all we needed from there? Just so we don't miss anything when adding new flavors.

PR #591, opened 2021-03-24
zstd compression for backups

CrimsonVoid:
## Proposed Changes
  - support zstd compression for backups; on my machine compressing ~880 mb takes roughly 30 seconds when using zip, and 3 seconds with zstd. Since zstd isn't widely supported I opted to keep zip backups the default for people who want backup's that "just work"

## Checklist
I was only able to get it to build on macOS, but it seems to work fine there.
- [x] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux


casperstorm:
Thank you for this PR, @CrimsonVoid. 
I am a little confused about the fileformat it uses (`.zst`) since if i understand it correctly you use `tar` together with `zstd` which normally gives a `.tar.zst`? 

```sh
# compress
tar --zstd -cf foo.tar.zst foo/

# decompress
tar --zstd -xf foo
```

I am not that much into compression so I am just wondering. Man it is fast though, awesome job.

CrimsonVoid:
I originally did want to use `.tar.zst` but `Archive::try_from(dest)` would fail since file_stems are different for the two, and `.zst` was the easiest way to fix it at the time. Looking back at this, I think using `.tar.zst` would be the clearest option here; I'll try to have a fix out later today.

File stem differences: 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=585c510f3c91a5765e7a4e1516ad1e28

casperstorm:
I see. Thanks for clarifying.
Let me know how it goes or if you need assistance.
This will be a nice QOL change for people doing automatic backups. And always nice with options.

CrimsonVoid:
Not the best fix, but I think it should work well enough. I also updated `latest_backup`, and I'm not sure if the update to the `Added` subheading in CHANGELOG.md was intentional or not (looks like the other Added sections use `###`), but if so, let me know and I can change it back.

casperstorm:
> Not the best fix, but I think it should work well enough. 

I think it looks good, I see the problems and I think you solved it in a nice way.

> I also updated `latest_backup`, and I'm not sure if the update to the `Added` subheading in CHANGELOG.md was intentional or not (looks like the other Added sections use `###`), but if so, let me know and I can change it back.

That was a mistake from my side, thanks for catching.
Ill do some more testing tomorrow, and will merge it during the weekend. Thanks again for the contribution and let me know if you want to work on more features!

casperstorm:
Confirmed working on Windows as well. Unzipping with [PeaZip](https://peazip.github.io/).

PR #598, opened 2021-03-26
fix: clippy warning from rust 1.51.0

casperstorm:
Resolves clippy warning from running Ajour with Rust 1.51.0

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/crates/core/src/config/wow.rs
+++ b/crates/core/src/config/wow.rs
@@ -31,43 +31,45 @@ impl Default for Wow {
 pub enum Flavor {
     #[serde(alias = "retail", alias = "wow_retail")]
     Retail,
-    RetailPTR,
+    #[serde(alias = "RetailPTR")]
+    RetailPtr,
     RetailBeta,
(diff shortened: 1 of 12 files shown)


crates/core/src/config/wow.rs
```diff
@@ -31,54 +31,56 @@ impl Default for Wow {
 pub enum Flavor {
     #[serde(alias = "retail", alias = "wow_retail")]
     Retail,
-    RetailPTR,
+    #[serde(alias = "RetailPTR")]
+    RetailPtr,
```
casperstorm, line 35: I chose to follow the clippy rule [upper_case_acronyms](https://rust-lang.github.io/rust-clippy/master/#upper_case_acronyms). This meant i had to rename our PTR Flavors which essentially would break reading our config file because suddenly `RetailPTR` does not exists. I fixed this by adding a serde alias, eventually we can get rid of it but for now I think it's the right choice to follow the standard.

PR #612, opened 2021-04-02
fix: other flavor

casperstorm:
## Proposed Changes
  - Add a `other` variant to Flavor to support potential new flavors added to the catalog. 

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/crates/core/src/config/wow.rs
+++ b/crates/core/src/config/wow.rs
@@ -39,6 +39,8 @@ pub enum Flavor {
     #[serde(alias = "ClassicPTR")]
     ClassicPtr,
     ClassicBeta,
+    #[serde(other)]
+    Other,
 }
 
 impl Flavor {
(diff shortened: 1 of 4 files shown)


crates/core/src/config/wow.rs
```diff
@@ -84,6 +95,7 @@ impl Flavor {
             Flavor::Classic => "_classic_".to_owned(),
             Flavor::ClassicPtr => "_classic_ptr_".to_owned(),
             Flavor::ClassicBeta => "_classic_beta_".to_owned(),
+            _ => panic!("Unsupported flavor: {:?}", self),
```
petergp, line 98: Will these cases ever be hit in practice?

casperstorm: It is used to convert a `Flavor` into the folder name, and those places we use this is all tied to `Flavor::ALL` which we control: 

```rust
    pub const ALL: [Flavor; 6] = [
        Flavor::Retail,
        Flavor::RetailPtr,
        Flavor::RetailBeta,
        Flavor::Classic,
        Flavor::ClassicPtr,
        Flavor::ClassicBeta,
    ];
```

I am not sure if its better to return a string, rather than just panic.

casperstorm:
If we are controlling the `Flavor` from the catalog, then this PR might not be needed @petergp.
Only case which we could end up in a issue is then here:

https://github.com/ajour/ajour-catalog/blob/c2d00e8e83b528a995292a252990842fa58a9733/townlong_yak.sh#L21

casperstorm:
@tarkah do you have anything to say about this? I am a little unsure if we want it.

PR #614, opened 2021-04-06
skip catalog addons with unknown game flavor

tarkah:
Use in favor of #612

## Proposed Changes
  - Skip de serializing unknown flavors in GameVersion and filter out any catalog items that have an empty game_versions array.

## Checklist

- [ ] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/crates/core/src/catalog.rs
+++ b/crates/core/src/catalog.rs
@@ -30,7 +30,10 @@ async fn get_catalog_addons_from(
                 .get("etag")
                 .and_then(|h| h.to_str().map(String::from).ok());
 
-            Ok(Some((etag, response.json::<Vec<CatalogAddon>>().await?)))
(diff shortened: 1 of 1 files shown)


crates/core/src/catalog.rs
```diff
@@ -117,7 +120,7 @@ pub struct CatalogAddon {
     #[serde(deserialize_with = "null_to_default::deserialize")]
     #[deprecated(since = "0.4.4", note = "Please use game_versions instead")]
     pub flavors: Vec<Flavor>,
```
casperstorm, line 122: Should we remove `flavors` all together now? It's been deprecated for a long time.

casperstorm:
LGTM, @tarkah.

PR #615, opened 2021-04-07
add sys tray support for windows

tarkah:
Resolves our biggest request, sys-tray for windows!

## Proposed Changes
  - New setting "Close to System Tray" for Windows
  - When this is selected, a system tray icon will be created.
  - Pressing close will hide Ajour. It can be opened by double clicking the tray icon, or right clicking tray icon -> show
  - When enabled, Ajour can only be closed via right click tray icon -> exit (we can't hook / override the Minimize button w/ Iced but as of 0.3, we can with the Close icon).
 

## Checklist

- [x] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/src/tray/mod.rs
+++ b/src/tray/mod.rs
@@ -20,9 +20,9 @@ use winapi::um::winuser::{
     RegisterClassExW, SendMessageW, SetFocus, SetForegroundWindow, SetMenuDefaultItem,
     SetWindowLongPtrW, ShowWindow, TrackPopupMenu, TranslateMessage, CREATESTRUCTW, GWLP_USERDATA,
     MAKEINTRESOURCEW, MB_ICONINFORMATION, MB_OK, MF_BYPOSITION, MF_GRAYED, MF_SEPARATOR, MF_STRING,
-    SW_HIDE, SW_RESTORE, SW_SHOWMINIMIZED, TPM_LEFTALIGN, TPM_NONOTIFY, TPM_RETURNCMD,
-    TPM_RIGHTBUTTON, WM_APP, WM_CLOSE, WM_COMMAND, WM_CREATE, WM_DESTROY, WM_INITMENUPOPUP,
-    WM_LBUTTONDBLCLK, WM_RBUTTONUP, WNDCLASSEXW, WS_EX_NOACTIVATE,
+    SW_HIDE, SW_RESTORE, TPM_LEFTALIGN, TPM_NONOTIFY, TPM_RETURNCMD, TPM_RIGHTBUTTON, WM_APP,
+    WM_CLOSE, WM_COMMAND, WM_CREATE, WM_DESTROY, WM_INITMENUPOPUP, WM_LBUTTONDBLCLK, WM_RBUTTONUP,
+    WNDCLASSEXW, WS_EX_NOACTIVATE,
 };
 
 use crate::log_error;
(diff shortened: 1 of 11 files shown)


src/tray/mod.rs
```diff
@@ -0,0 +1,424 @@
...
+            TranslateMessage(&msg);
+            DispatchMessageW(&msg);
+        }
+
+        if tray_state.close_gui {
+            // WM_QUIT was sent, which was triggered by the Exit button. Close entire program.
+            SHOULD_EXIT.store(true, Ordering::Relaxed);
+
+            // Activate window to force event loop to run / see that it should now close
+            ShowWindow(tray_state.gui_handle.unwrap(), SW_SHOWMINIMIZED);
```
casperstorm, line 166: Can this `.unwrap()` fail

src/tray/mod.rs
```diff
@@ -0,0 +1,424 @@
...
+    Shell_NotifyIconW(NIM_DELETE, &mut icon_data);
+}
+
+unsafe fn show_popup_menu(hwnd: HWND, state: &TrayState) {
+    let menu = CreatePopupMenu();
+
+    let hidden = state.gui_hidden;
+
+    let mut about = str_to_wide!("About...");
+    let mut toggle = str_to_wide!(if hidden { "Show window" } else { "Hide window" });
```
casperstorm, line 228: We I think we should order and name the `popup_menu` like:

```
Show Ajour
About
----------
Exit
```
Also wondering if theres an API to make a horizontal line as seen on a few trays?

image

src/tray/mod.rs
```diff
@@ -0,0 +1,424 @@
...
+    if msg == WM_CREATE {
+        let create_struct = &*(lparam as *const CREATESTRUCTW);
+        state = &mut *(create_struct.lpCreateParams as *mut TrayState);
+        SetWindowLongPtrW(hwnd, GWLP_USERDATA, state as *mut _ as LPARAM);
+
+        add_icon(hwnd);
+
+        if state.show_balloon {
+            display_balloon_message(hwnd, "Running from Tray...");
+        }
```
casperstorm, line 312: Where is this shown 🤔

casperstorm (commented):
Woah the win32 api is ... crazy. Not sure I can fully wrap my head around everything.

src/tray/mod.rs
tarkah: Good idea! Let me look into this

src/tray/mod.rs
tarkah: Nope, it's safe. Lines 117-119 is where we find the gui handle and assign it to the state before creating our window. So within the windows loop, it'll always be Some otherwise we would have never made it to the point of making the window.

tarkah: See the `enum_proc` procedure and the https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-enumwindows documentation.

Since Iced doesn't give us a way to reference the underlying window handle (HWND), I have to use this function to find it. Basically it enumerates over all windows and I just find the "active" window that has the same process id as Ajour, lol.

tarkah: We need that window handle so we can show / hide it w/

```rust
let cmd = if state.gui_hidden {
                    SW_RESTORE
                } else {
                    SW_HIDE
                };

                ShowWindow(*state.gui_handle.as_ref().unwrap(), cmd)
```

tarkah:
@casperstorm this includes the patched iced that rolls back wgpu. With merging this, we can close out #616. The rollback of wgpu in iced seems to fix both the resize crash and higher memory usage on windows.

PR #620, opened 2021-04-14
add theme import feature

tarkah:
Adds import functionality for themes created from theme builder

## Checklist

- [x] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/src/gui/element/settings.rs
+++ b/src/gui/element/settings.rs
@@ -13,7 +13,7 @@ use {
     ajour_core::{config::Config, theme::ColorPalette},
     iced::{
         button, pick_list, scrollable, Align, Button, Checkbox, Column, Container, Element, Length,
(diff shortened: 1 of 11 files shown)


src/gui/element/settings.rs
```diff
@@ -161,6 +161,67 @@ pub fn data_container<'a, 'b>(
...
+
+        let theme_input = TextInput::new(
+            &mut theme_state.input_state,
+            "Paste URL here...",
+            &theme_state.input_url,
+            Interaction::ThemeUrlInput,
+        )
+        .size(DEFAULT_FONT_SIZE)
+        .padding(6)
+        .width(Length::Units(300))
```
casperstorm, line 177: I think this looks better and should be enough!

```suggestion
        .width(Length::Units(185))
```

casperstorm (approved):
It's awesome. Great job mate.

PR #629, opened 2021-04-23
feat: burning crusade classic

casperstorm:
This PR will add support for Burning Crusade.

Currently we do not have enough information from Blizzard to know which naming they are choosing for the Burning Crusade, so until this is known, this PR will be blocked:

Mostly a TODO for myself.

- [x] Add _classic_beta_
  - [x] Hub
  - [x] Curse
- [x] Classic will be renamed from _classic_ to _classic_era_
- [x] Burning Crusade will use _classic_
  - [x] Hub
  - [x] Curse
  - [x] Create it as a new base flavor  

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux


fleps:
Does the current last release support TBC Beta in any way?
I've installed and I can see my TBC as "Classic Beta" on the path, but the catalog / install from Curse still uses / install the Classic version of the addons, not the TBC.

casperstorm:
> Does the current last release support TBC Beta in any way?
> 
> I've installed and I can see my TBC as "Classic Beta" on the path, but the catalog / install from Curse still uses / install the Classic version of the addons, not the TBC.

Not yet. We are awaiting Blizzard to expose folder structure to us. Also Curse API needs to update which has still hot happened.

joaomcarlos:
Looking forward to this. Everyones waiting :D

crates/core/src/config/wow.rs
```diff
@@ -34,18 +34,21 @@ pub enum Flavor {
     #[serde(alias = "RetailPTR")]
     RetailPtr,
     RetailBeta,
-    #[serde(alias = "classic", alias = "wow_classic")]
+    #[serde(rename = "classic", rename = "wow_classic", rename = "classic_era")]
```
casperstorm, line 37: @tarkah seems like we can't multiple `rename`.

PR #630, opened 2021-04-26
Feat/avoid multiple instances

### LINKED ISSUE #624: Multible instance of Ajour in systemtray
If you choose to let Ajour start to system tray automatically when windows starts and then start ajour manually, you get several Ajour instances in the systemtray

Ajour 1.1.0
Windows 10 20H2, OS Version: 19042.928

tarkah:
Resolves #624

## Proposed Changes
  - Exit if another instance of Ajour is running (on Windows)
  - Ajour will now write to a `pid` file the PID of the process and name of the executable.
  - When Ajour launches, it'll check the `pid` file and see if a process for that PID is running from that same executable name.
  - If so, the current process will exit
  - If not, the current process will overwrite the `pid` file with it's own info and continue to run

## Checklist

- [x] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/src/process.rs
+++ b/src/process.rs
@@ -1,0 +1,89 @@
+use std::ffi::c_void;
+use std::path::PathBuf;
+
+use ajour_core::fs::PersistentData;
+use serde::{Deserialize, Serialize};
+use winapi::{
+    shared::winerror::WAIT_TIMEOUT,
+    um::{
+        processthreadsapi::{GetCurrentProcess, GetCurrentProcessId, OpenProcess},
+        synchapi::WaitForSingleObject,
+        winbase::QueryFullProcessImageNameW,
+        winnt::{PROCESS_QUERY_LIMITED_INFORMATION, SYNCHRONIZE},
+    },
+};
+
+#[derive(Debug, Serialize, Deserialize)]
+struct Process {
+    pid: u32,
+    name: String,
+}
+
+impl PersistentData for Process {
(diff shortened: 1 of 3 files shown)


src/process.rs
```diff
@@ -0,0 +1,96 @@
...
+    }
+}
+
+pub fn avoid_multiple_instances() {
+    if process_already_running() {
+        log::info!("Another instance of Ajour is already running. Exiting...");
+        std::process::exit(0);
+    } else {
+        // Otherwise this is only instance. Save info about this process to the
+        // pid file so future launches of Ajour can detect this running process.
```
casperstorm, line 34:
```suggestion
        // Otherwise this is the only instance. Save info about this process to the
        // pid file so future launches of Ajour can detect this running process.
```

src/process.rs
```diff
@@ -0,0 +1,96 @@
...
+struct Process {
+    pid: u32,
+    name: String,
+}
+
+impl PersistentData for Process {
+    fn relative_path() -> PathBuf {
+        PathBuf::from("pid")
+    }
+}
```
casperstorm, line 26: Smart move to use `PersistentData` :)

casperstorm (approved):
Really nice, I think this is the right way to go and makes a lot of sense.

src/process.rs
tarkah: Best trait ever :)

PR #644, opened 2021-05-22
Feature - Backup screenshots

Cronnay:
## Proposed Changes
  - Ability to backup screenshots
  - Refactor "get_x_directory" to reuse code, also easier to add new one.

## Checklist

- [x] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/crates/core/src/config/mod.rs
+++ b/crates/core/src/config/mod.rs
@@ -3,7 +3,7 @@ use crate::error::FilesystemError;
 use crate::repository::CompressionFormat;
 use glob::MatchOptions;
 use serde::{Deserialize, Serialize};
-use std::{collections::HashMap};
+use std::collections::HashMap;
 use std::fmt::{self, Display, Formatter};
 use std::fs::create_dir_all;
 use std::path::{Path, PathBuf};
@@ -159,7 +159,11 @@ impl Config {
                         ..Default::default()
                     };
 
-                    let pattern = format!("{}/{}", wow_flavor_dir.display(), get_pattern_format(relative_path));
+                    let pattern = format!(
+                        "{}/{}",
+                        wow_flavor_dir.display(),
+                        get_pattern_format(relative_path)
+                    );
 
                     for path in glob::glob_with(&pattern, options).unwrap().flatten() {
                         dir = path;
@@ -367,8 +371,14 @@ mod test {
 
     #[test]
     fn test_get_format_interface_addons() {
-        assert_eq!(get_pattern_format("Interface/Addons"), String::from("?nterface/?ddons"));
-        assert_eq!(get_pattern_format("iNtErFaCe/aDdoNs"), String::from("?nterface/?ddons"));
+        assert_eq!(
+            get_pattern_format("Interface/Addons"),
+            String::from("?nterface/?ddons")
+        );
+        assert_eq!(
+            get_pattern_format("iNtErFaCe/aDdoNs"),
+            String::from("?nterface/?ddons")
+        );
     }
 
     #[test]
@@ -379,11 +389,19 @@ mod test {
         assert_eq!(get_pattern_format("wTf"), String::from("?tf"));
     }
 
-    
     #[test]
     fn test_get_format_screenshots() {
-        assert_eq!(get_pattern_format("Screenshots"), String::from("?creenshots"));
-        assert_eq!(get_pattern_format("sCREENSHOTS"), String::from("?creenshots"));
-        assert_eq!(get_pattern_format("ScreeNShots"), String::from("?creenshots"));
+        assert_eq!(
+            get_pattern_format("Screenshots"),
+            String::from("?creenshots")
+        );
+        assert_eq!(
+            get_pattern_format("sCREENSHOTS"),
+            String::from("?creenshots")
+        );
+        assert_eq!(
+            get_pattern_format("ScreeNShots"),
+            String::from("?creenshots")
+        );
     }
 }
--- a/crates/core/src/config/mod.rs
+++ b/crates/core/src/config/mod.rs
@@ -3,7 +3,7 @@ use crate::error::FilesystemError;
 use crate::repository::CompressionFormat;
 use glob::MatchOptions;
 use serde::{Deserialize, Serialize};
-use std::collections::HashMap;
+use std::{collections::HashMap};
 use std::fmt::{self, Display, Formatter};
 use std::fs::create_dir_all;
 use std::path::{Path, PathBuf};
@@ -44,6 +44,9 @@ pub struct Config {
 
     #[serde(default)]
     pub backup_config: bool,
+
+    #[serde(default)]
+    pub backup_screenshots: bool,
 
     #[serde(default)]
     pub hide_ignored_addons: bool,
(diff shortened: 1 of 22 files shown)


crates/core/src/config/mod.rs
```diff
@@ -164,36 +134,62 @@ impl Config {
...
+
+    /// Returns a `Option<PathBuf>` to the Screenshots directory.
+    /// This will return `None` if no `wow_directory` is set in the config.
+    pub fn get_screenshots_directory_for_flavor(&self, flavor: &Flavor) -> Option<PathBuf> {
+        self.get_directory_for_flavor(flavor, "Screenshots")
+    }
+
+    /// Return a `Option<PathBuf` to a directory.
+    /// This will return none if no `wow_directory` is set in the config.
+    fn get_directory_for_flavor(&self, flavor: &Flavor, relative_path: &str) -> Option<PathBuf> {
```
casperstorm, line 148: I like this. Great to introduce some reusable code. Good job.

crates/core/src/config/mod.rs
```diff
@@ -164,36 +134,62 @@ impl Config {
...
+                if !dir.exists() {
                     let options = MatchOptions {
                         case_sensitive: false,
                         ..Default::default()
                     };
 
                     // For some reason the case insensitive pattern doesn't work
                     // unless we add an actual pattern symbol, hence the `?`.
-                    let pattern = format!("{}/?tf", dir.display());
+                    let pattern = format!("{}/{}", wow_flavor_dir.display(), get_pattern_format(relative_path));
```
casperstorm, line 164: We should update the comment, now that you have refactored the "issue" into `get_pattern_format`.

crates/core/src/config/mod.rs
```diff
@@ -164,36 +134,62 @@ impl Config {
...
+fn get_pattern_format(relative_path: &str) -> String {
+    let mut splitted_string = relative_path.split("/");
+    let mut return_string: Vec<String> = vec![];
+    while let Some(path) = splitted_string.next() {
+        let mut to_lower_case = path.to_lowercase();
+        to_lower_case.replace_range(0..1, "?");
+        return_string.push(to_lower_case);
+    }
+    return_string.join("/")
+}
```
casperstorm, line 191: We should add a comment to this function.

crates/core/src/config/mod.rs
```diff
@@ -364,3 +360,29 @@ pub async fn load_config() -> Result<Config, FilesystemError> {
 const fn default_true() -> bool {
     true
 }
+#[cfg(test)]
+mod test {
```
casperstorm, line 364: Good job adding tests!!

casperstorm (changes requested):
@Cronnay great to see you contribution, welcome to the team! Such a great first addition.

crates/core/src/config/mod.rs
```diff
@@ -164,36 +134,62 @@ impl Config {
...
+
+    /// Return a `Option<PathBuf` to a directory.
+    /// This will return none if no `wow_directory` is set in the config.
+    fn get_directory_for_flavor(&self, flavor: &Flavor, relative_path: &str) -> Option<PathBuf> {
+        let wow_flavor_dir = self.wow.directories.get(flavor);
+        match wow_flavor_dir {
+            Some(wow_flavor_dir) => {
+                let mut dir = wow_flavor_dir.join(relative_path);
 
                 // If path doesn't exist, it could have been modified by the user.
```
casperstorm, line 154:
```suggestion
                // If dir doesn't exist, it could have been modified by the user.
```
Small nitpick now we use the word `dir `instead of `path`

crates/core/src/config/mod.rs
```diff
@@ -164,36 +134,62 @@ impl Config {
...
+
+                Some(dir)
             }
             None => None,
         }
     }
 }
 
+fn get_pattern_format(relative_path: &str) -> String {
+    let mut splitted_string = relative_path.split("/");
```
casperstorm, line 183:
```suggestion
    let mut splitted_string = relative_path.split('/');
```

I think this could be a char instead.

crates/core/src/config/mod.rs
```diff
@@ -164,36 +134,62 @@ impl Config {
...
+                }
+
+                Some(dir)
             }
             None => None,
         }
     }
 }
 
+fn get_pattern_format(relative_path: &str) -> String {
```
casperstorm, line 182: We should add a comment to this function.

crates/core/src/config/mod.rs
casperstorm, line 191: I think we can rewrite this function, and avoid the `while `and a `mut`:

```rust
fn get_pattern_format(relative_path: &str) -> String {
    let splitted_string = relative_path.split('/');
    let mut return_string: Vec<String> = vec![];
    for path in splitted_string {
        let mut to_lower_case = path.to_lowercase();
        to_lower_case.replace_range(0..1, "?");
        return_string.push(to_lower_case);
    }
    return_string.join("/")
}
```

crates/core/src/config/mod.rs
Cronnay: Actually moved comment to the method instead

crates/core/src/config/mod.rs
Cronnay: Fixed

crates/core/src/config/mod.rs
Cronnay: Yes, I agree! Had issue to figure out how to handle split correctly. Thank you

Fixed.

crates/core/src/config/mod.rs
Cronnay: Took the former comment regarding the reason for the question mark, also comment for intellisense.

crates/core/src/config/mod.rs
```diff
@@ -3,7 +3,7 @@ use crate::error::FilesystemError;
 use crate::repository::CompressionFormat;
 use glob::MatchOptions;
 use serde::{Deserialize, Serialize};
-use std::collections::HashMap;
+use std::{collections::HashMap};
```
tarkah, line 6: Not sure how the `{  }` remained here. Can you remove or run `cargo fmt`?

src/cli.rs
```diff
@@ -148,6 +149,7 @@ fn str_to_backup_folder(s: &str) -> Result<BackupFolder, &'static str> {
         "wtf" => Ok(BackupFolder::Wtf),
         "addons" => Ok(BackupFolder::AddOns),
         "config" => Ok(BackupFolder::Config),
-        _ => Err("valid values are ['all','wtf','addons','config']"),
+        "screenshots" => Ok(BackupFolder::Screenshots),
+        _ => Err("valid values are ['all','wtf','addons','config', 'screenshots']"),
```
tarkah, line 153: Can you also add `screenshots` to the structopt description for field `backup_folder`... Line 102

src/command/backup.rs
```diff
@@ -1,4 +1,4 @@
-use crate::cli::BackupFolder;
+use crate::{cli::BackupFolder};
```
tarkah, line 1: Same here... remove `{  }`

src/command/backup.rs
```diff
@@ -61,8 +61,14 @@ pub fn backup(
                 None => continue,
             };
 
+            let screenshot_directory = match config.get_screenshots_directory_for_flavor(&flavor) {
+                Some(path) => path,
+                None => continue,
```
tarkah, line 66: Does the screenshot folder always exist? If not, we shouldn't `continue` here unless `ALL` or `screenshot` was passed on the command line. Otherwise it'll cause backup to fail even when the user doesn't want to backup the screenshots folder, if they don't have screenshots.

tarkah (changes requested):
@Cronnay Thanks for your first contribution! It looks good! Just a few small things that should be addressed.

src/cli.rs
tarkah: This is what prints when an error occurs (user passes the wrong text). The description from line 102 is what prints on `--help`

src/command/backup.rs
Cronnay: Installed PTR to make sure, it does not always exist. The current implementation will create a screenshots directory however. It wasnt a problem when I tried to backup non-existing directory.

Maybe not a decent solution, what could be a better implementation? Maybe initialize in said task. Example below (sorry about poor indentation lol)

```rust
match backup_folder {
      ....
      BackupFolder::Screenshots => {
        if let Some(screenshot_directory) = config.get_screenshots_directory_for_flavor(&flavor) {
          let screenshots_folder = backup::BackupFolder::new(&screenshot_directory, &wow_directory); 
          if screenshot_directory.exists() {
           src_folders.push(screenshots_folder);
          }
        };
    }
}
```

then do something similar for the "ALL" task

crates/core/src/config/mod.rs
Cronnay: You are correct - Not sure why it added curly braces. Made a commit only for formatting.

src/cli.rs
Cronnay: Added!

src/command/backup.rs
tarkah: Ahh yes, my bad. The function returns None only if the wow directory isn't set for that flavor, not if that folder doesn't exist. 

And you already have a check on `if_exists` before pushing the backup folder. So I think this is actually fine!

Cronnay: Cheers! then I'm resolving it :)

tarkah (approved):
Thanks for the changes!! Looks good to me.

casperstorm (approved):
Great contribution man.

PR #646, opened 2021-05-23
feat(locale): add polish locale

KasprzykM:
## Proposed Changes
  - Adds polish translations to Ajour.

## Checklist

- [x] Tested on Windows
- [ ] Tested on MacOS ---> Unable to test that one.
- [ ] Tested on Linux ---> Haven't yet.
- [x] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users ---> Wasn't sure if I should do that for something this small.


casperstorm (changes requested):
Looks good @KasprzykM. All future language changes is done through https://poeditor.com/join/project?hash=gjghTR7d0Z, just FYI. Also, there seems to be a crash when polish is selected and you go to Catalog. Can you verify this? It seems to be crashing here:

```rust
let isolang = isolang::Language::from_locale(&lang).unwrap();
```

I have a feeling its because `csb_PL` is not supported by `isolang` so a `unwrap` is not great here. Maybe you can take a look?

KasprzykM:
Woops, you were completely right. I've used the wrong code, not sure why it has not crashed for me before though. It is fixed now and I confirmed that isoland properly detects that as `Pol`.

Was not aware of poeditor - my bad:)

casperstorm:
> Woops, you were completely right. I've used the wrong code

Good catch!

> Was not aware of poeditor - my bad:)

All good! Just took your translation and uploaded it. Just changes going forward will go through POEditor!

KasprzykM:
Added in `Unreleased` section to leave decision about release up to you guys :)

casperstorm:
Thanks @KasprzykM ❤️

PR #661, opened 2021-06-20
Add Polish pl translation

LEGOL2:
It's worth to mention that Polish verbs change form, depending on context, so this translation might not be 100% accurate. I will update it if I find something off.

## Proposed Changes
  - Add Polish PL translation

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux


casperstorm:
Thanks for the contribution, but we do have a polish translation ready on the development branch:
See: https://github.com/ajour/ajour/blob/development/locale/pl.json.

If you want to help contribute with it, then we do it through POEditor: https://poeditor.com/join/project?hash=gjghTR7d0Z

PR #670, opened 2021-07-02
Fixed old theme link (I think)

seank-com:
Resolves #

## Proposed Changes
  - Found a link to a deleted THEMES.md updated to scroll down to themes section instead.

## Checklist

- [x] Doc bug (no platform testing required)
- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux


tarkah:
Good catch!

Issue #660, opened 2021-06-15
[Bug] "Delete SavedVariables when deleting addons" has two issues.


Bananaman:
This ticket is for tracking two minor issues with this feature:

1. It removes `addon.lua` properly but leaves the `addon.lua.bak` files.
2. When addons use lots of folders, it doesn't properly delete the SavedVariables at all, from what I can see? Tried uninstalling Deadly Boss Mods Core (TBC) and it left lots of DBM-related savedvars behind. :-)

tarkah:
@Bananaman  I've fixed 1. with #671.

For 2. what are the names of some of the SavedVariables files the remain? The file _should_ match one of the addon folders of the addon that gets deleted.

Bananaman:
@tarkah Hi, apologies for not seeing your question sooner!

I can confirm that #671 fixes both bugs. The remnants of DBM were definitely .bak files without me noticing that at the time. I just did a re-install of DBM, ran the addon, let it make a dozen savedvar files, manually created .bak copies with a script, and then ran Ajour's uninstall: All cleaned up! :-)

PR #671, opened 2021-07-02
properly remove .lua and .lua.bak files

tarkah:
Resolves first issue in #660

## Proposed Changes
  - Use file_name over file_stem, then properly remove .lua and .lua.bak

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [x] Tested on Linux


casperstorm (approved):
Excellent idea with temp file! LGTM.

PR #678, opened 2021-07-10
feat: better ui for displaying errors to the user

### LINKED ISSUE #649: [Classic ERA] no addon detected
**Describe the bug**
Sometimes, Ajour don't find any addon (i have around 50 of them).
After some time, and multiple closing/reopening Ajour, it works again.

**Screenshots**
Untitled0
Untitled

**Software involved**
 - OS: Windows 10
 - Ajour version: 1.2.0
 - Classic ERA

casperstorm:
Resolves #649

## Proposed Changes
  - When encountering an error, it is now possible to attach it to the current mode, which will then display it on the screen rather than in the little error box top right. This means, e.g., in the case of problems reaching the host, we write that rather than "No addons found". See the picture below as an example. I've chosen to keep the error box top right because it is still used for updating, unpacking, hashing errors.

Screenshot 2021-07-10 at 22 15 34

- We don't have `State::Start` anymore, this made little sense anyway. We now have `Ready` `Loading` and `Error`.

## Checklist

- [ ] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/src/gui/mod.rs
+++ b/src/gui/mod.rs
@@ -43,15 +43,14 @@ static WINDOW_ICON: &[u8] = include_bytes!("../../resources/windows/ajour.ico");
 
 #[derive(Debug)]
 pub enum State {
-    Start,
     Ready,
     Loading,
-    Error(Option<anyhow::Error>),
+    Error(anyhow::Error),
 }
 
 impl Default for State {
     fn default() -> Self {
-        State::Start
+        State::Ready
     }
 }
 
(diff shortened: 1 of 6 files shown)


casperstorm:
For now, I've only replaced `ajour.error` for the three modes: `MyAddons`, `MyWeakAuras`, and `Catalog`. I think this is a good start.

src/gui/mod.rs
```diff
@@ -1043,26 +1036,38 @@ impl Application for Ajour {
...
+                            &error_description,
+                            None,
+                        ))
+                    }
+                    None => Some(element::status::data_container(
+                        color_palette,
+                        &localized_string("setup-ajour-title")[..],
+                        &localized_string("setup-ajour-description")[..],
+                        Some(&mut self.onboarding_directory_btn_state),
+                    )),
```
casperstorm, line 1060: Before we used `State::Start` to know if it was "first time" but I figured it was better just to check if we had anything in our state HashMap and utilize the fact that if there's nothing in it, it must mean we are in an "Start" state.

PR #682, opened 2021-07-12
feat: zstd compression slider

### LINKED ISSUE #672: Add Zstd Custom Compression Level Option
The backup compression / option for Zstd uses the default compression level (default: 3). With my specific backup folder, which is 83.9MiB, the zstd compression will create a 17.1MiB file. This is pretty fantastic given it performs this compression in 300ms, however I would like to request that the ability to set our own compression level. In terms of the UX, I imagine this option could be exposed in the most straightforward manner with some form of a compression slider. I threw a mockup together in Inkscape to illustrate it.

ajour-compression-slider

Here are some metrics I ran for some higher zstd compression levels. This is an average of 5 runs.

Compression Level | Time | Resulting File Size
---- | ---- | ----
-8 | 1.158s | 15.8MiB
-13 | 4.946s | 15.4MiB
-18 | 17.872s | 14.4MiB
-19 | 27.04s | 14.1MiB
-22 (with --ultra) | 45.878s | 13.8MiB

The zstd rust package provides a mechanism to update the `CompressionLevel` for the encoder via [set_parameter](https://docs.rs/zstd/0.9.0+zstd.1.5.0/zstd/stream/write/struct.Encoder.html#method.set_parameter).

Assuming the implementation of the mockup above, I would propose:

- Fastest: 3 (default)
- Faster: 8
- Fast: 13
- Slow: 18
- Slowest: 22

casperstorm:
Resolves #672

## Proposed Changes
  - ~~Essentially I've extended `CompressionFormat::Zstd` to `CompressionFormat::Zstd(i32)` to hold a compression level.~~
  - ~~I've had to make a deserde migration to convert `CompressionFormat::Zstd => CompressionFormat::Zstd(i32)`~~
  - Added a new field to store `zstd_compression_level`.
  - Added gui slider to handle the selection for compression level.

## Checklist

- [ ] Tested on Windows
- [x] Tested on MacOS
- [x] Tested on Linux


crates/core/src/config/mod.rs
```diff
@@ -181,6 +181,49 @@ impl Config {
     }
 }
 
+pub mod compression_format {
```
casperstorm, line 184: This part feels a little "too much," so if you have any improvements here @tarkah, please let me know.
Essentially I have extended `CompressionFormat::Zstd` to take an `i32` now, so I need a migration from `Zstd` -> `Zstd(i32)` in order not to break config.

What I have now works, but it just feels like a lot of code for such a small change—damn migrations.

src/main.rs
```diff
@@ -85,7 +85,7 @@ pub fn main() {
                     destination,
                     flavors,
                     compression_format,
-                } => command::backup(backup_folder, destination, flavors, compression_format),
+                } => command::backup(backup_folder, destination, flavors, compression_format, 0),
```
casperstorm, line 88: For CLI i just use default (fastest).

JoshStrobl:
Tested on Linux and confirmed to work as intended. Choosing various slider options for Zstd and clicking to backup correctly generates zstd compressed tarballs of varying sizes:

Some examples:

```
-rw-rw-r--  1 joshua joshua 17800482 Jul 14 15:28 ajour_backup_2021-07-14_15-28-51.tar.zst
-rw-rw-r--  1 joshua joshua 16968562 Jul 14 15:28 ajour_backup_2021-07-14_15-28-58.tar.zst
-rw-rw-r--  1 joshua joshua 16450511 Jul 14 15:29 ajour_backup_2021-07-14_15-29-07.tar.zst
-rw-rw-r--  1 joshua joshua 14362747 Jul 14 15:29 ajour_backup_2021-07-14_15-29-11.tar.zst
```

Switching back and forth between zip and zstd correctly toggles the visibility of the slider.

Huge thanks for this excellent work.

crates/core/src/fs/backup.rs
```diff
@@ -11,7 +11,7 @@ use zip::{write::FileOptions, CompressionMethod, ZipWriter};
 
 /// A trait defining a way to back things up to the fs
 pub trait Backup {
-    fn backup(&self) -> Result<()>;
+    fn backup(&self, level: Option<i32>) -> Result<()>;
```
tarkah, line 14: Should `level` be a `ZstdBackup` field instead of requiring it in the trait method, since not all trait impls use `level`, like `ZipBackup`?

tarkah (commented):
Looks great! I think adding the new `level` field over encoding it in `CompressionFormat` is much cleaner.

crates/core/src/fs/backup.rs
casperstorm: Yes, good point. I don't know why I didn't think about that. It has no purpose in a generic trait like that. It has been removed now.

Silun:
The 'slowest' option takes a _really_ long time to finish. I'd say it does impractically so, but putting that issue aside - Ajour will use a single thread that takes up a full CPU core for the compression. Have you thought about making Ajour spawn a different thread to do the compression, and having that thread be low priority? This way Ajour itself would still work just fine, and the compression won't be slowing down whatever else one does on the PC. I'd assume somebody who is using an explicitly slow setting is not going to care about the couple of seconds that the priority will add to the duration of the compression.

Anyway, a neat feature. I like having control about the compression setting :)

PR #684, opened 2021-07-13
feat: font backup

casperstorm:
Resolves #675 

## Proposed Changes
  - Ability to backup Fonts folder.

Should be merged after #682.

## Checklist

- [ ] Tested on Windows
- [X] Tested on MacOS
- [x] Tested on Linux
- [X] Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users

PR #685, opened 2021-07-13
feat: delete saved variables from addon

### LINKED ISSUE #658: [Feature] Deleting SavedVariables on-demand for individual addons
It's very frustrating to delete orphaned SavedVariables manually, or when doing it manually to reset specific addon settings.

Ajour has a "delete SavedVariables when uninstalling" feature but it's too heavy-handed, because I could forget that it's on and just delete/re-install something and lose all my settings.

I propose a secondary feature:

1. Click on an addon in My Library.
2. In the expanded view with "Exclude", "Delete" etc, have a "Wipe SavedVariables" button.
3. Clicking the button must bring up a dialog "Are you sure you want to delete all settings for addon X?" Yes/No, to avoid accidents.

This was the *single nice feature* I saw during my brief use of CurseForge. Their client has a right-click "Delete SavedVariables" menu entry and it was super useful.

casperstorm:
Resolves #658 

## Proposed Changes
  - Added a new button to delete SavedVariables on a single addon.
  - Added a new confirmation flow where you have to press twice on a "delete" button in order to confirm it.

## Checklist

- [ ] Tested on Windows
- [X] Tested on MacOS
- [ ] Tested on Linux


JoshStrobl:
Validated on Linux. Worth noting that the locale/en.json changes will need to be rebased after #682 and #684 lands, otherwise it'll fail to cleanly apply. Ended up removing both of those PRs from my local build, added this patch and continued validation.

As a sidenote, probably would've been a good idea for me to double check that I had `WTF` checked in my backups before I did a "Delete Saved Variables" during testing. But hey, can't fix stupid :D

Bananaman:
@casperstorm This feature works very well and I like the GUI a lot. Great job! Tried it with addons that consist of a lot of folders and a ton of separate var files and all got deleted when I cleared that addon's savedvars. :)

Bananaman:
Oh and I just realized another great use of this new feature. It lets me keep the heavy-handed "always delete savedvars when deleting addons" disabled in the options. And when I uninstall something, I can instead just click that addon, click "delete its savedvars" and then "delete this addon". All in a neat and easily usable GUI. Awesome. :)

PR #687, opened 2021-07-14
feat: added support for release.json

### LINKED ISSUE #686: Ajour won't respect release.json when installing from GitHub
**Describe the bug**
When installing from GitHub, Ajour may install the wrong version, e.g. of this repo:

https://github.com/raethkcj/RatingBuster

It has a release file that should tell Ajour what to install, but it still installs the wrong version by default for Classic TBC at least.

This is the content of the `release.json`:

```json
{
    "releases": [{
        "filename": "RatingBuster-v1.0.12-bcc.zip",
        "nolib": false,
        "metadata": [{
            "flavor": "bcc",
            "interface": 20501
        }]
    }, {
        "filename": "RatingBuster-v1.0.12-classic.zip",
        "nolib": false,
        "metadata": [{
            "flavor": "classic",
            "interface": 11307
        }]
    }]
}
```

**To Reproduce**
1. Install Ratingbuster for TBC CLassic via Ajour from `https://github.com/raethkcj/RatingBuster`
2. Check the version it has installed in the toc file
3. See that it has installed version 11307

**Expected behavior**
It should have installed the version indicated for 20501

**Software involved**
 - OS: Windows 10
 - Ajour version: 1.2.2
 - TBC Classic

**Additional context**

**Log Output**
Log of updating the addon today, same issue applies.
[ajour.log](https://github.com/ajour/ajour/files/6816657/ajour.log)

casperstorm:
Resolves #686 

## Proposed Changes
  - Ajour will now properly respect `release.json` when installing through Github and Gitlab.
    - `release.json` is auto-generated when using https://github.com/BigWigsMods/packager.
  - Also corrected a small bug which caused our (now fallback) solution to not work properly with `-bcc` assets.
 
Test URLs with `release.json`:
  * https://github.com/WeakAuras/WeakAuras2
  * https://github.com/Hekili/hekili
  * https://github.com/raethkcj/RatingBuster

## Checklist

- [x] Implement same change for Gitlab
- [ ] Tested on Windows
- [X] Tested on MacOS
- [ ] Tested on Linux


crates/core/src/repository/backend/git.rs
```diff
@@ -193,70 +226,78 @@ mod github {
...
+                .find(|r| r.metadata.iter().any(|m| m.flavor == flavor))
+            {
+                // Find the asset, based on what we know from the `release.json`.
+                let asset = release
+                    .assets
+                    .iter()
+                    .find(|a| a.name == release_file.filename);
+                Ok(asset)
+            } else {
+                Ok(None)
```
tarkah, line 256: We should fall back to the "fallback solution" here

crates/core/src/repository/backend/git.rs
```diff
@@ -193,70 +226,78 @@ mod github {
...
+            }
+        } else {
+            // Fallback solution where we try to look at the asset name to determine the correct file.
+            // Eg:
+            // `foobar-classic` => ClassicEra file.
+            // `foobar-bcc` => ClassicTbc file.
```
tarkah, line 262: This fallback solution has one downside compared to the old method... For classic / bcc, if no `classic` or `bcc` named file exists, respectively, but a single `.zip` asset exists, it won't use that.

tarkah: Most authors should conform to the new `release.json` or multi-asset naming convention, so I'm fine taking a stance against this and just informing authors of how to fix their releases.

tarkah (approved):
Looks good at a glance from me! There is just one potential "regression" that I outlined, though I'm fine if we take the stance to ignore it.

crates/core/src/repository/backend/git.rs
casperstorm: Agree, all repos which is using BigWigs packager will get this file, so I would assume it will become defacto standard.

crates/core/src/repository/backend/git.rs
casperstorm: This is now fixed. Good catch.

PR #692, opened 2021-07-17
feat: keybindings

casperstorm:
Resolves https://discord.com/channels/757155234500968459/761096457238478898/850545677356236814

| key        | action                 |
|------------|------------------------|
|  <kbd>a</kbd>       | Go to MyAddons         |
| <kbd>w</kbd>          | Go to Wago             |
| <kbd>c</kbd>          | Go to Catalog          |
| <kbd>i</kbd>          | Go to Install from URL |
| <kbd>s</kbd>          | Go to Settings         |
| <kbd>r</kbd>          | Refresh                |
| <kbd>u</kbd>          | Update                 |
| <kbd>Tab</kbd>        | Go to next tab         |
| <kbd>Ctrl</kbd> + <kbd>Tab</kbd>| Go to previous tab     |

This is just some basic key bindings, what others should we add?

### Ideas:
- Shortcut to Ajour config folder
- Shortcut to Wow folder
- ?

## Checklist

- [ ] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux
- [x] Added an entry to `README.md`


casperstorm:
@tarkah could be cool to do:
`tab` - cycle modes.
`ctrl+tab` cycle modes backward.

tarkah:
Yeah I think tab is definitely the way to go

Bananaman:
These look absolutely great!

And as I mentioned in a discord discussion:

- . Literally the only thing I hope for with `c` for catalog is that it auto-focuses the searchbox to make it easy to search instantly. Instead of pressing `c` and then grabbing the mouse to move to and click the field.
- Shortcut to Wow folder: Yes, I use this all the time! Maybe something like `Shift-W` for "WTF"?
- Shortcut to Ajour config folder: Sure why not, but it's not something I'd use. Maybe `Shift-A`?

Reason for Shift: Making it harder to open those folders by accident. And Ctrl is already taken on all operating systems.

casperstorm:
> * Shortcut to Wow folder: Yes, I use this all the time! It would be a huge help to be able to open it via Ajour! Maybe something like `Shift-W` for "WTF"/"Warcraft"?
> 
> * Shortcut to Ajour config folder: Sure why not, but it's not something I'd use. Maybe `Shift-A`?

I like `Shift-W` and `Shift-A`.

casperstorm:
On MacOS it is common that settings/preferences is <kbd>CMD</kbd>+<kbd>,</kbd>.
I think we should stick with  <kbd>s</kbd> as **S**ettings and keep it like that on all platforms.

Thoughts on that?

PR #693, opened 2021-07-17
fix: check flavors for font folder

casperstorm:
We now check all flavors for any font folder.

## Checklist

- [ ] Tested on Windows
- [X] Tested on MacOS
- [ ] Tested on Linux
- [ ] ~~Added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users~~


tarkah (approved):
Lgtm!

PR #694, opened 2021-07-18
fix: parsing regexes coverted to raw strings

casperstorm:
## Proposed Changes
  - Changes relevant regexes to raw regexes (Thanks @Bananaman)

Note: this needs to be proper tested.

## Checklist

- [ ] Tested on Windows
- [x] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ and `Removed`.
 
 ## [Unreleased]
 
+### Changed
+
+- Speed up parsing addons by improved regex expessions.
+
 ## [1.2.3] - 2021-07-17
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and `Removed`.
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


casperstorm:
@Bananaman Should we change these as well?

https://github.com/ajour/ajour/blob/924386eb67d52e45092d1d0d52ebb27fe01085ab/crates/core/src/parse.rs#L1015-L1016

https://github.com/ajour/ajour/blob/924386eb67d52e45092d1d0d52ebb27fe01085ab/crates/core/src/parse.rs#L1024-L1025

Bananaman:
Yeah, let's make every regex raw to clean things up. :)

Bananaman:
@casperstorm +1. After carefully reading through every change, it all looks correct. Functionality of every regex is the same before and after, but it's a lot cleaner and easier to edit now.

If it all passes through testing/usage (which it should), it's good to go now!

casperstorm:
@Bananaman Great, thanks again.
@tarkah if you have time, could you test this PR and see if you have any troubles with install/matching addons?

PR #695, opened 2021-07-22
feat: add hub catalog source

casperstorm:
Since Townlong Yak has been [deprecated](https://www.townlong-yak.com/addons/about/update-api) I've chosen to remove it from the Catalog, but instead added Hub support. Hub is a new a upcoming repository from WowUp.

Depends on https://github.com/ajour/catalog/pull/2

## Proposed Changes
  - Remove Townlong Yak and add support for Hub (since we used Hub for TLY, it is mainly a renaming task).
  - Added a `All` source when viewing Catalog.

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux


crates/core/src/repository/mod.rs
```diff
@@ -13,15 +13,16 @@ use std::str::FromStr;
...
+use backend::{Curse, Github, Gitlab, Hub, Tukui, WowI};
 
 #[derive(Debug, Clone, Copy, PartialEq, PartialOrd, Ord, Eq, Serialize, Deserialize)]
 pub enum RepositoryKind {
     Curse,
     Tukui,
     WowI,
-    TownlongYak,
+    #[serde(alias = "TownlongYak")]
+    Hub,
```
casperstorm, line 25: We can do this to make sure that addons installed from TownlongYak will continue and use Hub going forward. This is because we used TLY with Hub as a backend before, so they are plug and play.

If we remove it, the addons will fall back to Curse. This will actually still work because TLY addons recently were added to Curse.

@tarkah Which migration solution do you like the best? Rely on fallback or add the alias.

crates/core/src/catalog.rs
```diff
@@ -6,7 +6,8 @@ use chrono::prelude::*;
 use isahc::AsyncReadResponseExt;
 use serde::{Deserialize, Serialize};
 
-const CATALOG_URL: &str = "https://raw.githubusercontent.com/ajour/catalog/main/catalog-0.1.0.json";
+const CATALOG_URL: &str =
+    "https://raw.githubusercontent.com/ajour/catalog/feat/hub-source/catalog-0.1.0.json";
```
casperstorm, line 10: TODO: This has to be changed before we release it.

crates/core/src/config/mod.rs
```diff
@@ -91,6 +91,15 @@ pub struct Config {
     pub start_closed_to_tray: bool,
 }
 
+/// Deserialize `catalog::Source`.
+/// If we find a unknown type, we return None rather than panic.
+fn from_catalog_source<'de, D>(deserializer: D) -> Result<Option<catalog::Source>, D::Error>
```
casperstorm, line 96: I've made a custom deserializer, because if the user had selected Townlong Yak as the `catalog::source` it would fail when loading `Config`. Now, it just returns `None` if it does not know the enum type.

crates/core/src/catalog.rs
```diff
@@ -6,7 +6,8 @@ use chrono::prelude::*;
 use isahc::AsyncReadResponseExt;
 use serde::{Deserialize, Serialize};
 
-const CATALOG_URL: &str = "https://raw.githubusercontent.com/ajour/catalog/main/catalog-0.1.0.json";
+const CATALOG_URL: &str =
+    "https://raw.githubusercontent.com/ajour/catalog/feat/hub-source/catalog-0.2.0.json";
```
casperstorm, line 10: This has to be changed.

crates/core/src/config/mod.rs
tarkah: You could also alias source like you did for RepoKind and it'll just load that one?

crates/core/src/repository/mod.rs
tarkah: I like the alias idea, since that's the source those add-ons are hooked up to anyway 👍

tarkah (commented):
Looks good! I still need to test out, but should all work since it's truly just a rename. And opening up catalog to return all add-ons vs just TY add-ons should *just work*

crates/core/src/config/mod.rs
casperstorm: I think it's better to keep it the same so I'll do alias as you suggest!

PR #696, opened 2021-07-22
feat: combined catalog source

casperstorm:
Quite a few people are annoyed that it's not possible to just see a combined list of addons in the Catalog.

## Proposed Changes
  - (Re)added the "All" source.

## Checklist

- [ ] Tested on Windows
- [X] Tested on MacOS
- [ ] Tested on Linux


src/gui/update.rs
```diff
@@ -1833,14 +1833,18 @@ pub fn handle_message(ajour: &mut Ajour, message: Message) -> Result<Command<Mes
...
-            log::debug!("Interaction::CatalogResultSizeSelected({:?})", source);
+            log::debug!("Interaction::CatalogSourceSelected({:?})", source);
 
             // Catalog source
             if let CatalogSource::Choice(source) = source {
                 ajour.config.catalog_source = Some(source);
-                let _ = ajour.config.save();
+            } else {
+                ajour.config.catalog_source = None;
             }
```
tarkah, line 1843: Just to confirm, if they select ALL we set the config to None because we unwrap none back as ALL? Makes sense!

tarkah (approved):
Looks good! Funny how we came full circle on this. I'll test later tonight.

src/gui/update.rs
casperstorm: Correct, I'll however update this with a little more explanation because it might look a little strange.

PR #701, opened 2021-07-24
Feature/addon import export

tarkah:
Resolves: https://github.com/ajour/ajour/issues/659

Initial commit, this only adds the proposed data structure for exporting and a couple functions

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux


crates/core/src/share.rs
```diff
@@ -0,0 +1,92 @@
...
+            id,
+            name,
+        })
+    }
+}
+
+pub fn export(
+    addons: HashMap<Flavor, Vec<Addon>>,
+    output_dir: impl AsRef<Path>,
+) -> Result<(), error::FilesystemError> {
```
tarkah, line 36: With this function, we can easily just pass a cloned map of the addons and it'll save it to the supplied path as "ajour-addons.yml"

crates/core/src/share.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    fs::write(file, contents)?;
+
+    Ok(())
+}
+
+pub fn parse_only_needed(
+    existing_addons: HashMap<Flavor, Vec<Addon>>,
+    import_string: String,
+) -> Result<HashMap<Flavor, Vec<Data>>, serde_yaml::Error> {
```
tarkah, line 63: There is no "import" function because I envision us just hooking right into the existing "install" process on the gui update loop. So this just parses the file and removes collisions. The resulting output has all the data needed to hook right into what `Message::Interaction(Interaction::InstallAddon` does.

tarkah: We should instead return a `Parsed` structure that has the hashmap, but also a field that tells how many collisions were removed so we can display this in gui

tarkah: I've added the Parsed structure in latest commit

tarkah:
I think we change the `output_dir` of export to be a full file path that the user selects via `natve-dialog`

crates/core/src/share.rs
```diff
@@ -0,0 +1,106 @@
...
+    output_dir: impl AsRef<Path>,
+) -> Result<(), error::FilesystemError> {
+    let data = addons
+        .into_iter()
+        .map(|(flavor, addons)| {
+            (
+                flavor,
+                addons
+                    .into_iter()
+                    .map(|a| Data::try_from((flavor, a)))
```
casperstorm, line 44: When we use `flavor` i feel like we should use `addon` rather than `a`.

```suggestion
                    .map(|addon| Data::try_from((flavor, addon)))
```

tarkah:
@casperstorm Import / export buttons are hooked up and work great! As a phase 2, it'd be nice to have a "Confirmation" screen with the parsed addons prior to installing, where we can display how many collisions were removed and even allow "deselecting" addons.

tarkah:
Another cool idea would be to add the export file to backups!

tarkah:
Also, we need to build out the CLI command for this functionality.

casperstorm:
Code looks great @tarkah.
However, the updated `native-dialog` on MacOS crashes:

```rust
fatal runtime error: Rust cannot catch foreign exceptions
[1]    39509 abort      cargo run
```

casperstorm:
This LGTM for a first iteration. After merge to master we update all translations.

PR #708, opened 2021-07-31
Feature/no self update

### LINKED ISSUE #704: Allow completely disabling update checking for those who use a package manager
Using a package manager the update button continue to shows up and an accidental click causes the program to update itself rather than through the package manager

There should be an option to completely disable update checks so that the update button does not show up at all

**Additional context**
Besides browsers and game clients every single program on my PC has the option to fully disable auto updates. For this reason I believe this is a reasonable request to have implemented not to mention it would take very little effort.

That being said thanks so much for this amazing addon manager!!

tarkah:
Resolves #704

## Proposed Changes
  - Adds a new cargo feature flag `no-self-update`. When used, Ajour won't check for newer updates.
  - Adds 2 new release assets, one for Windows and Macos, both with this feature activated. These can be picked up by Chocolatey & Homebrew respectively.

## Checklist

- [ ] Tested on Windows
- [x] Tested on MacOS
- [x] Tested on Linux

### DIFF
--- a/.github/workflows/release_to_draft.yml
+++ b/.github/workflows/release_to_draft.yml
@@ -4,11 +4,10 @@ on:
   workflow_dispatch:
     inputs:
       tag:
-        description: 'Specify tag to create'
+        description: "Specify tag to create"
         required: true
 
 jobs:
-  
   build:
     name: Build
     strategy:
@@ -21,6 +20,10 @@ jobs:
           - target: windows-opengl
             os: windows-latest
             make: make binary OPENGL=1
+            binary_path: target/release/ajour.exe
+          - target: windows-noselfupdate
+            os: windows-latest
+            make: make binary NOSELFUPDATE=1
             binary_path: target/release/ajour.exe
 
           - target: linux-wgpu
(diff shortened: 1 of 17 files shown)


tarkah:
@casperstorm Can you try `make tar MACOS=1 NOSELFUPDATE=1` and confirm it's built correctly? When running the binary, the `Message::LatestRelease(Some("1.2.5"))` should instead be `Message::LatestRelease(None)`. We no longer fire off a request to Github to check for releases.

tarkah:
I'm not adding a new asset for Linux, or updating the APPIMAGE to use this feature, since we still want users using APPIMAGE to get _notified_ when a new release is available. APPIMAGE doesn't have the ability to self update, but they still get a notice.

Linux packagers, like @joshstrobl, don't rely on a release asset and instead can compile using `cargo build --release --features no-self-update`.

casperstorm:
> @casperstorm Can you try `make tar MACOS=1 NOSELFUPDATE=1` and confirm it's built correctly? When running the binary, the `Message::LatestRelease(Some("1.2.5"))` should instead be `Message::LatestRelease(None)`. We no longer fire off a request to Github to check for releases.

Confirmed and working on macOS:

```
19:40:11.957 [ajour::gui::update][DEBUG] Message::LatestRelease(None)
```

casperstorm (approved):
LGTM!

kwilliams1987:
I will update the Chocolatey package to support this once 1.3 is released.

This might result in one version which has the auto update still enabled as the Chocolatey package is updated every 6 hours.

PR #709, opened 2021-07-31
Fix/tab no modifiers

tarkah:
Resolves #

## Proposed Changes
  - Alt-tabbing shouldn't cause Ajour to change tabs

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@ and `Removed`.
 - Adds 2 new release assets, one for Windows and Macos, both with this feature
   activated. These can be picked up by Chocolatey & Homebrew respectively.
 
+### Fixed
+
+- Fixed bug where `alt+tab` would change tabs in Ajour.
+
 ## [1.2.5] - 2021-07-26
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


casperstorm (approved):
LGTM! Also this tested this branch on macOS.

Neurotoxin001:
It's still not fixed even in 1.3.0
https://github.com/ajour/ajour/releases/tag/1.3.0

tarkah:
> It's still not fixed even in 1.3.0

It appears the issue is when alt-tabbing _back_ to Ajour. When alt-tabbing away from Ajour, the tab doesn't change. It seems the issue is when you alt-tab back to Ajour, and release the Tab key once Ajour is back in focus. This is the behavior on Windows, but the problem doesn't occur on Linux.

The bigger issue is we are capturing these hotkeys on key released, instead of key pressed. I believe switching over to key pressed should resolve this issue. @casperstorm any issues with making this change?

tarkah:
Hmm, changing to on key pressed actually makes this problem now occur on Linux as well.

PR #724, opened 2021-08-12
feat: retry buttons

### LINKED ISSUE #717: Add "Retry" button to retry requests when failing to download catalogs or addons

This scenario was reported on Discord where a user could experience a network timeout when downloading the catalog, though the problem likely extends to other networking requests such as add-on downloading, such as when the user is saturating their network connection when downloading an WoW update, PTR, etc.

At the moment, the only method for resolution is to restart Ajour. Ideally in this scenario (error code 6, "Couldn't resolve host name") a retry button should be presented to retry the network request.

No other alternatives have been considered.

**Additional context**

N/A

casperstorm:
Resolves #717 

## Proposed Changes
  - Adds a retry button under the error to try again.

Screenshot 2021-08-12 at 22 27 24

## Checklist

- [ ] Tested on Windows
- [X] Tested on MacOS
- [ ] Tested on Linux

### DIFF
--- a/src/gui/mod.rs
+++ b/src/gui/mod.rs
@@ -256,7 +256,6 @@ pub struct Ajour {
     backup_state: BackupState,
     column_settings: ColumnSettings,
     catalog_column_settings: CatalogColumnSettings,
-    onboarding_directory_btn_state: button::State,
     catalog: Option<Catalog>,
     install_addons: HashMap<Flavor, Vec<InstallAddon>>,
     catalog_last_updated: Option<DateTime<Utc>>,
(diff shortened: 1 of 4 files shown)


src/gui/mod.rs
```diff
@@ -350,6 +352,10 @@ impl Default for Ajour {
             pending_confirmation: None,
             zstd_compression_level_slider_state: Default::default(),
             share_state: Default::default(),
+            onboarding_status_button_state: Default::default(),
+            addons_status_button_state: Default::default(),
+            weakauras_status_button_state: Default::default(),
+            catalog_status_button_state: Default::default(),
```
tarkah, line 358: Does this have to be 4 buttons or can it be one? Since it's only passed once / on different pattern branches.

casperstorm: Good point. Fixed.

tarkah (approved):
LGTM!

PR #732, opened 2021-09-10
fix/elvui-tbc

casperstorm:
This fixes the issue where a few addons woudnt show up in Ajour because of flavor toc paths were changed.

### DIFF
--- a/crates/core/src/parse.rs
+++ b/crates/core/src/parse.rs
@@ -274,26 +274,29 @@ async fn parse_addon_folders(
         .filter_map(|id| {
             // Generate .toc path.
             let toc_path = {
-                let multi_part = match flavor.base_flavor() {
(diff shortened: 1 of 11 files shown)


crates/core/src/parse.rs
```diff
@@ -272,24 +272,31 @@ async fn parse_addon_folders(
...
+                        Flavor::ClassicEra => vec!["Classic", "Vanilla"],
+                        Flavor::ClassicTbc => vec!["BCC", "TBC"],
+                        _ => vec![],
+                    };
+
+                    for part in multi_part {
+                        for separator in &["-", "_"] {
+                            let toc_with_flavor = root_dir
+                                .join(&id)
+                                .join(format!("{}{}{}.toc", id, separator, part));
```
tarkah, line 289: Good idea! Check both ways just in case.

PR #734, opened 2021-09-11
Add support for Classic Era PTR

jncl:
Resolves #

## Proposed Changes
  -

## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [ ] Tested on Linux


casperstorm:
Oh that is a thing now? Ill look at adding support for it.

casperstorm:
Didn't know this was a thing :)
Thanks for adding the support. I've tested and confirmed this is working.

PR #735, opened 2021-09-12
Only package as opengl and make default for building

tarkah:
## Checklist

- [ ] Tested on Windows
- [ ] Tested on MacOS
- [x] Tested on Linux

### DIFF
--- a/.github/workflows/release_to_draft.yml
+++ b/.github/workflows/release_to_draft.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         target:
-          - target: windows-wgpu
+          - target: windows
             os: windows-latest
             make: make binary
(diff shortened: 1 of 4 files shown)


tarkah:
@casperstorm This makes `opengl` the default for building Ajour. It updates our CI to only publish `opengl` releases. For backwards compatibility, we are still releasing the `-opengl` builds for the self-update path, but they are identical now to the normal builds, both are opengl.

The self-update path has been fixed to ALWAYS download the non-opengl releases. This will allow us to remove the `-opengl` releases in the release after this one. Note this would technically break older `opengl` clients from being able to self-update, so we can always leave the `-opengl` builds for backwards compatibility. Your call.

casperstorm (approved):

